PR #4, opened 2025-04-13
fix: correct URL parameter handling in fetchLibraryDocumentation

thomasleveil:
This PR fixes an issue with URL parameter handling in the `fetchLibraryDocumentation` function. The changes:

1. Replace incorrect `contextURL.set()` with `contextURL.searchParams.set()` for proper URL parameter handling
2. Convert tokens to string when setting as URL parameter
3. Use consistent double quotes for string literals


enesgules (approved):
Thanks!

Issue #7, opened 2025-04-14
Recommend deno instead of NPX.


ryanleecode:
MCP servers should use deno instead of node.js because node.js is a very insecure runtime. If there is ever a MCP supply chain attack users will be at risk!

The readme should recommend deno with these settings!

```json
{
  "mcpServers": {
    "context7": {
      "command": "deno",
      "args": ["run", "--allow-net", "npm:@upstash/context7-mcp"]
    }
  }
}

```

enesgules:
Most MCP servers and libraries use Node, and that’s what the MCP docs use too. Curious to hear more on why you think Deno would be better.

ryanleecode:
> Most MCP servers and libraries use Node, and that’s what the MCP docs use too. Curious to hear more on why you think Deno would be better.

Just because the majority does it doesnt mean its better. With deno's runtime you can control what permission it has. 

In this case by setting only allow net it means it will not have access to your file system or environment variables

enesgules:
The MCP server in this case doesn’t really deal with the file system or env variables — it just sends requests to a remote API. So there’s not much risk in practice.

Node is also the default runtime in MCP docs and is more widely supported across platforms and tooling, so sticking with Node keeps things simpler and more consistent for most users.

Thanks again for the suggestion!

ryanleecode:
Yes it doesnt do X until a malicious pereon changes the code and uploads it to npm without somebody noticing.

this isnt asking the code to be change, simply updating the readme since deno can run npm packages as shown in the description of the issue

enesgules:
Can you please send a PR?

vhbui02:
> MCP servers should use deno instead of node.js because node.js is a very insecure runtime. If there is ever a MCP supply chain attack users will be at risk!
> 
> The readme should recommend deno with these settings!
> 
> {
>   "mcpServers": {
>     "context7": {
>       "command": "deno",
>       "args": ["run", "--allow-net", "npm:@upstash/context7-mcp"]
>     }
>   }
> }

Exactly what I want. Personally I use npm but inside a Rootless Docker container with read-only permission and limited access to host filesystem.

PR #21, opened 2025-04-19
chore: update readme to recommend deno instead of npx

ryanleecode:
[#7](https://github.com/upstash/context7-mcp/issues/7)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ How do I invalidate a query in React Query? use context7
 How do I protect a route with NextAuth? use context7
 ```
 
-Context7 fetches up-to-date documentation and working code examples right into your LLM’s context.
+Context7 fetches up-to-date documentation and working code examples right into your LLM's context.
 
 - 1️⃣ Ask your question naturally
 - 2️⃣ Tell the LLM to `use context7`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,8 +45,24 @@ No tab-switching, no hallucinated APIs that don't exist, no outdated code genera
 
 Go to: `Settings` -> `Cursor Settings` -> `MCP` -> `Add new global MCP server`
 
+Using `deno` is recommended for enhanced security, as it allows explicit control over script permissions (like network access) via flags like `--allow-net`, reducing potential risks compared to runtimes with broader default permissions. See [#7](https://github.com/upstash/context7-mcp/issues/7) for discussion.
```
enesgules, line 48: Could you please remove this part, we don't want to confuse people and we still want to keep npx as default

enesgules (changes requested):
Could you please add deno as alternative, we still want to keep npx as default.

ryanleecode:
> Could you please add deno as alternative, we still want to keep npx as default.

I think that would defeat the purpose as 90% of people will not open the alternative box. deno should be the default because it will keep users SAFU.

But I will respect your decision as the maintainer so please let me know

enesgules:
This was the final decision after discussing it with the team. Thank you, and sorry for the inconvenience.

ryanleecode:
done

enesgules (approved):
Thanks!

PR #24, opened 2025-04-19
chore: Update README.md to support Windows MCP configuration

SahilPanhotra:
Added Windows specific configuration for cursor, windsurf and vs code


enesgules (changes requested):
Do we really need to add it multiple times? Maybe only one in the end is enough? Or just in the Cursor section?

PR #70, opened 2025-04-23
docs: add README-zh-CN and add a button in README.md

duyifan02:
Hello! I really like your work, and it should be promoted to more people. I have added a Chinese README document for you, which can help more people understand this project faster.


enesgules:
Thank you so much!

enesgules (approved):

PR #74, opened 2025-04-23
[Translation] Korean README

coyaSONG:
- Add Korean translation (README.ko.md) in /docs folder
- Add Korean badge to main README.md (I applied it in green, but do I need to do it in yellow?)


enesgules (approved):
Thank you so much! Color is not important.

PR #77, opened 2025-04-23
Add VS Code install buttons

digitarald:
Adds a nice one-click instal experience, similar to what the GitHub and Playwright servers offer in their readme.


enesgules (approved):
Thank you so much!

PR #84, opened 2025-04-24
[Translation] Spanish README

dsfaccini:
I kept the original "use context7" as the other two translations did as well. "use" in spanish means "[you] use" in a formal context, the direect translation of "use" would be "usa". I didn't wanna confuse the next translators so I'll stick to "use".


enesgules (approved):
Thank you so much!

PR #95, opened 2025-04-25
Dynamic Versioning and Dependency Update

ksp2000:
### Summary
This pull request introduces the following changes:

- **Feature:** Implements dynamic versioning for the Context7 server.
- **Chore:** Bumps the version to `1.0.7` in `package.json` and `version.ts`.
- **Dependency Update:** Updates `@modelcontextprotocol/sdk` dependency to version `1.10.2`.

### Details
- The dynamic versioning ensures that the server version is automatically updated based on the latest changes.
- The dependency update aligns the project with the latest features and fixes provided by the SDK.

### Testing
- Verified that the versioning logic works as expected.
- Ensured compatibility with the updated SDK version.

### Impact
- Improves maintainability and reduces manual effort in version management.
- Leverages the latest improvements in the SDK for enhanced functionality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@upstash/context7-mcp",
-  "version": "1.0.6",
+  "version": "1.0.7",
   "description": "MCP server for Context7",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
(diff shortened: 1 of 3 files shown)


enesgules (commented):
Thank you so much for this.
However, we are actually releasing the package using GitHub Actions, and we don’t use the version from package.json. Instead, we use the Git tag for versioning.

This PR actually gave me the idea to update the MCP server version through the release action as well, so thanks for that! I have now updated the release action file: https://github.com/upstash/context7/blob/master/.github/workflows/release.yml

PR #98, opened 2025-04-25
Fix for The recent "[ERR_MODULE_NOT_FOUND]: Cannot find module '...@modelcontextprotocol/sdk/dist/esm/shared/uriTemplate.js" Bug in Various IDE 

JyotirmoyDas05:
# Context7 MCP Server ESM Module Resolution Fix

## Issue Summary

The Context7 MCP server was failing to start when launched via `npx` with the following error:

> Error: Cannot find module 'uriTemplate.js'

This error prevented users from installing and using the MCP server through the standard installation instructions provided in the `README`.

---

## Root Cause Analysis

The issue stems from ESM (ECMAScript Modules) resolution problems:

- The `@modelcontextprotocol/sdk` package uses ES modules with internal imports.
- When installed via `npx` in a temporary directory, Node.js cannot correctly resolve internal module paths.
- The specific missing file `uriTemplate.js` is referenced by other modules but can't be located.
- This issue is particularly problematic with newer Node.js versions that have stricter ESM path resolution.

---

## Solution Implemented

The issue has been resolved by adding the `--node-options=--experimental-vm-modules` flag to the Node.js execution:

- This flag enables enhanced ESM support in Node.js.
- It allows proper resolution of dependencies and enables the server to start successfully.

---

## Changes Made

- Updated the installation instructions in the `README.md` file for all supported platforms:
  - Cursor
  - Windsurf
  - VS Code
  - Claude Desktop
- Each section now includes the working configuration with the necessary flag.
- Pinned the version to `@1.0.6` instead of `@latest` for consistency.
- Retained all alternative installation methods (Bun, Deno) as backup options.
- Added notes about using the flag in the troubleshooting section.

---

## Testing

The solution has been verified in multiple environments:

- ✅ Successfully runs in Cursor IDE  
- ✅ Successfully runs in VS Code on Windows  
- ✅ Successfullu runs in Claude Desktop
- ✅ Sucessfully runs in Windsurf IDE

---

## Recommendation for Long-Term Fix

To improve long-term maintainability, consider the following approaches:

- Bundle all dependencies to avoid dynamic ESM resolution issues.
- Update the package to fully support ESM without requiring experimental flags.
- Add a pre-check script that verifies Node.js compatibility before launching.
- Update package.json to explicitly include all required dependencies with compatible versions.
- Convert critical paths to CommonJS or ensure full ESM compatibility

---

## Related Documentation

There is a known issue with Node.js ESM resolution that this fix addresses.  
The `--experimental-vm-modules` flag is documented in the [Node.js documentation](https://nodejs.org/api/cli.html#--experimental-vm-modules).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,11 @@ Pasting the following configuration into your Cursor `~/.cursor/mcp.json` file i
   "mcpServers": {
     "context7": {
       "command": "npx",
-      "args": ["-y", "@upstash/context7-mcp@latest"]
+      "args": [
+        "-y", 
+        "--node-options=--experimental-vm-modules",
+        "@upstash/context7-mcp@1.0.6"
+      ]
     }
   }
 }
@@ -107,7 +111,11 @@ Add this to your Windsurf MCP config file. See [Windsurf MCP docs](https://docs.
   "mcpServers": {
     "context7": {
       "command": "npx",
-      "args": ["-y", "@upstash/context7-mcp@latest"]
+      "args": [
+        "-y", 
+        "--node-options=--experimental-vm-modules",
+        "@upstash/context7-mcp@1.0.6"
+      ]
     }
   }
 }
@@ -126,7 +134,11 @@ Add this to your VS Code MCP config file. See [VS Code MCP docs](https://code.vi
     "Context7": {
       "type": "stdio",
       "command": "npx",
-      "args": ["-y", "@upstash/context7-mcp@latest"]
+      "args": [
+        "-y", 
+        "--node-options=--experimental-vm-modules",
+        "@upstash/context7-mcp@1.0.6"
+      ]
     }
   }
 }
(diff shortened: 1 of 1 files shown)


JyotirmoyDas05:
I am a Amateur Developer So I might have done some Mistakes Here and There which Might Have Gone Unnoticed. I would Kindly Like you to review the Changes in the **Readme.md** file and asses any Security Concern You might think the Changes made may have.

Alternatively Try to Fix the Issue Presented included in The Long-Term Fix section of my Description

I loved your Project as It was Helping me from Copy/Pasting Links from Various Documentation for a Simple Fix in Cursor IDE So I tried to Help out when Encountered with this error and saw many other in the Issue Section complaining about it. Hope it may help you

enesgules (changes requested):
Thanks a lot for the detailed investigation!
The default config works fine for most users, so we’d prefer not to add extra flags by default.
However, we can definitely add your solution under the Troubleshooting section in the README for users who encounter ESM resolution issues.
Could you update the PR to do that instead?

JyotirmoyDas05:
> Thanks a lot for the detailed investigation!
> The default config works fine for most users, so we’d prefer not to add extra flags by default.
> However, we can definitely add your solution under the Troubleshooting section in the README for users who encounter ESM resolution issues.
> Could you update the PR to do that instead?

Thanks for the response 

I will Surely do That then

JyotirmoyDas05 (commented):
I have added the troubleshooting codes in the Troubleshooting section to address potential issues users might encounter. Please review and let me know if any additional scenarios need to be covered or if more adjustments are required.

enesgules (approved):
Thank you so much!

PR #115, opened 2025-04-27
fix: correct badge links in README.md for VS Code installation

luantranminh:
There is a redundant `config` key in the configuration, which prevents VSCode from starting the MCP server.

image

This PR will remove it.

image


enesgules (approved):
Tysm!

PR #132, opened 2025-04-30
[Translation] German README

dsfaccini:
Add the german translation of the current README. Following the same principles as with the pull request for the spanish translation.


enesgules (approved):
Thank you so much!

Issue #19, opened 2025-04-18
getting error


fisforfaheem:
![Image](https://github.com/user-attachments/assets/a03c62e1-5012-40b8-8d75-2dfca1d153b6)

kamiljaneczek:
+1 have the same issue

JvdsReform:
+1, I have the same error.

R3verseIN:
I have the same error.

klppl:
Running Cursor (Linux) also get "client closed".

```
2025-04-19 09:52:04.679 [info] >>>>> MCP Startup
2025-04-19 10:50:01.444 [info] ext7: Handling CreateClient action
2025-04-19 10:50:01.445 [info] ext7: Starting new stdio process with command: npx -y @upstash/context7-mcp@latest
2025-04-19 10:50:01.445 [error] ext7: Client error for command A system error occurred (spawn npx ENOENT)
2025-04-19 10:50:01.445 [error] ext7: Error in MCP: A system error occurred (spawn npx ENOENT)
2025-04-19 10:50:01.446 [info] ext7: Client closed for command
2025-04-19 10:50:01.447 [error] ext7: Error in MCP: Client closed
2025-04-19 10:50:01.470 [info] ext7: Handling CreateClient action
2025-04-19 10:50:01.470 [info] ext7: Starting new stdio process with command: npx -y @upstash/context7-mcp@latest
2025-04-19 10:50:01.470 [error] ext7: Client error for command A system error occurred (spawn npx ENOENT)
2025-04-19 10:50:01.470 [error] ext7: Error in MCP: A system error occurred (spawn npx ENOENT)
2025-04-19 10:50:01.471 [info] ext7: Client closed for command
2025-04-19 10:50:01.471 [error] ext7: Error in MCP: Client closed
2025-04-19 10:50:01.473 [info] ext7: Handling ListOfferings action
2025-04-19 10:50:01.473 [error] ext7: No server info found
2025-04-19 10:50:01.473 [info] ext7: Handling ListOfferings action
2025-04-19 10:50:01.473 [error] ext7: No server info found
2025-04-19 10:50:12.836 [info] ext7: Handling ReloadClient action
2025-04-19 10:50:12.836 [info] ext7: Starting new stdio process with command: npx -y @upstash/context7-mcp@latest
2025-04-19 10:50:12.837 [error] ext7: Client error for command A system error occurred (spawn npx ENOENT)
[...]

PR #166, opened 2025-05-07
docs: Adding Notes on Using MCP in Windows Systems

citizenll:
Solution for issue #19 on Windows, updated the Chinese and default README.


enesgules:
Maybe we can reformat this and create a separate section called "install in Windows"? Currently there is no such section. wdyt? @citizenll

enesgules (approved):
Thank you so much!

PR #182, opened 2025-05-12
Add docker compose file and shell script for easy local deployment via Docker

jitsmaster:
Please let me know if I need to provide window powershell or batch file scripts too.


enesgules:
Why did you remove the bun.lock file and added package.lock?

enesgules (commented):
Hey, we want to keep using bun that's why I am closing this PR, create a new one please for the docker compose file

Issue #210, opened 2025-05-20
stop using environment variables


ryanleecode:
No reason that we should be using environment variables and dotenv; this is breaking deno as the only permission i want to pass is `allow-net`. the MCP server shouldnt require access to my file system.

We can easily just pass in command line arguments.

Contextually when I raised this issue, the response was

> "The MCP server in this case doesn’t really deal with the file system or env variables — it just sends requests to a remote API. So there’s not much risk in practice."

https://github.com/upstash/context7/issues/7#issuecomment-2815118700

ryanleecode:
The alternative approach would be to keep using that environment variable but not using `dotenv` as it will try to search your file system for `.env` files -- we shouldn't require access to the filesystem as per the discussion in https://github.com/upstash/context7/issues/7

PR #208, opened 2025-05-19
added sse http support

buggyhunter:

PR #211, opened 2025-05-20
chore: use cli arguments instead of environment variables

ryanleecode:
https://github.com/upstash/context7/issues/210


mdogan:
I think we should support both (env and cli args) not to break existing installations. Args can have priority over env (if exists). Otherwise we can fallback to env vars as it's now.

ryanleecode:
> I think we should support both (env and cli args) not to break existing installations. Args can have priority over env (if exists). Otherwise we can fallback to env vars as it's now.

Can we agree to remove dotenv if that is the case? This mcp server should not have access to the file system

mdogan:
> > I think we should support both (env and cli args) not to break existing installations. Args can have priority over env (if exists). Otherwise we can fallback to env vars as it's now.
> 
> Can we agree to remove dotenv if that is the case? This mcp server should not have access to the file system

I agree. @enesakar & @enesgules: do you have any objections?

enesgules:
There might be new env variables with the new sse and http support pr: https://github.com/upstash/context7/pull/208 
Let’s revisit this once that pr is merged

ryanleecode:
> There might be new env variables with the new sse and http support pr: https://github.com/upstash/context7/pull/208 
> Let’s revisit this once that pr is merged

if we are adding new env variables we should prefix them with `CONTEXT7_` or something equivalent. this makes it easy to whitelist via `--allow-env=CONTEXT7_*`

enesgules:
> if we are adding new env variables we should prefix them with `CONTEXT7_` or something equivalent. this makes it easy to whitelist via `--allow-env=CONTEXT7_*`

Hey @ryanleecode there are new env variables added thanks to the new transport protocols that we support, do you think adding the prefix is enough or should we still support cli args. If so would you mind closing this PR since it's a bit old and creating a new one?

enesgules:
Closing this since we have a new pr https://github.com/upstash/context7/pull/278

PR #248, opened 2025-06-01
Update lockfile + simple ci github check

vaporif:
Looks like lockfile wasn't updated for some time.
I've added ci check so this won't happen again


enesgules (approved):
TYSM!

PR #255, opened 2025-06-03
docs: add installation instructions for Roo Code MCP integration

likweitan:
Verified connection with streamable-http and stdio, both works in Roo Code.


enesgules (approved):
Thanks!

PR #291, opened 2025-06-13
chore: update deno readme

ryanleecode:
There's no reason to allow broad based access to all environment variables. As of right now, there is only the downstream dependency depd (https://github.com/dougwilson/nodejs-depd) which require the NO_DEPRECATION, TRACE_DEPRECATION. Lets limit the env variables to just this alone.

Also it advisible to setup some sort of CI to ensure these always work.


enesgules (approved):
Thanks!

PR #292, opened 2025-06-13
feat: Add accept-list and reject-list repo filters

### LINKED ISSUE #247: Feature Request: Context7 Console should allow for acceptList and rejectList as an input for MCP
list of repos to be allowed or rejected

akbxr:
Closes #247 

This PR introduces new filtering capabilities to the CLI tool for repository selection and updates the documentation accordingly. It adds support for `--accept-list` and `--reject-list` options, enabling users to include or exclude repositories based on patterns. Additionally, it integrates these filters into the search results processing logic and enhances the user experience with detailed feedback about applied filters.

| Before | After |
|---|---|
| Screenshot 2025-06-13 at 22 08 49 | Screenshot 2025-06-13 at 22 04 42 |

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,11 +539,64 @@ bun run dist/index.js
 
 - `--transport <stdio|http|sse>` – Transport to use (`stdio` by default).
 - `--port <number>` – Port to listen on when using `http` or `sse` transport (default `3000`).
+- `--accept-list <repos...>` – Only include repositories matching these patterns (supports wildcards: `*` and `?`).
+- `--reject-list <repos...>` – Exclude repositories matching these patterns (supports wildcards: `*` and `?`).
 
-Example with http transport and port 8080:
+#### Examples
 
+Basic usage with http transport and port 8080:
 ```bash
 bun run dist/index.js --transport http --port 8080
+```
+
+Only include Next.js repository:
+```bash
+bun run dist/index.js --accept-list vercel/next.js
+```
+
+Include only repositories from Vercel and TanStack:
+```bash
+bun run dist/index.js --accept-list vercel/* tanstack/*
+```
+
+Exclude deprecated repositories:
+```bash
+bun run dist/index.js --reject-list */deprecated-*
+```
+
+Combine accept and reject lists:
+```bash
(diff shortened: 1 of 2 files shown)


enesgules:
Thanks for your contribution. We’re still reviewing it and thinking about how we can further improve the developer experience of this feature. So it might take a bit of time before we can merge this PR.

kylesnowschwartz:
I'd like to add my input that this is a highly desired feature that is currently preventing my team from adopting this tool. The fear is that we may inadvertently supply the AI with instructions to install or perform a destructive action (prompt injection attacks). By limiting what Context7 can look up with a defined list of approved repos we would limit that attack vector.

Thanks for your hard work @enesgules !

enesgules:
We felt that a better solution was possible and we are working on it atm. Please let us know what you think @kylesnowschwartz 

image

kylesnowschwartz:
It's great that you're working on a robust solution, thank you. I'll add a few of my thoughts:

- In addition to a GUI, being able to pass these as arguments to the MCP server via the JSON config for a cli-first solution would be (more) convenient
- The list of documentation types from Github repos would ideally also be configurable for full flexibility
- 'Documentation Websites' remains pretty opaque. Maybe there's a way to make that more specific or discoverable what constitutes a documentation website?
- More verbose tool tips for full-text explanations of what adding a trusted org does (does it bypass the toggles above? Does it include all files or just the 'allowed-list' of file types? etc.). In fact just more verbose tool tips I think are best, i.e. "explain it like I'm 5" principle.

I'll share this around my team as well to see if there are any thoughts but this looks like great progress, thank you for engaging and supporting AI tooling! Cheers.

enesgules:
this is now available through context7.com/dashboard 

we have decided to move with this approach to keep the MCP and endpoints fairly simple. If we want to implement a feature such as this, we would have to either filter on the client side, which is slow and wouldn't work for every use case since the endpoint returns a limited number of results. or we have to send the filters to the server which would complicate the endpoints. 

thanks for your patience and your contributions especially @akbxr <3

PR #294, opened 2025-06-14
Add instructions for AugmentCode

IslandApps:
Updated documentation to include the correct AugmentCode instructions.


IslandApps:
Please double check, I didn't mean to change any of the existing instructions and meant to only add my instructions. The reason I did this is because when you follow the original instructions before my changes, you're required to figure out more on your own, or it's possible I wasn't following them correctly. Regardless, the instructions I provided seem to be the easiest way with the latest version of Augment.

enesgules (approved):
Thank you!

PR #334, opened 2025-06-25
docs: fix typo

w2xi:
The correct term for simplified Chinese is "简体中文", which is written in simplified characters.

On the other hand, "簡體中文" is the traditional Chinese version of the same term.


enesgules (approved):
xiexie!

PR #337, opened 2025-06-25
docs: add Gemini CLI installation instructions

Ifan24:


enesgules (approved):
Wow that was fast! Thank you so much!

PR #398, opened 2025-07-11
docs: add remote server install instructions for Gemini CLI

tahatariq19:
Gemini CLI added support for remote MCP servers via Custom HTTP headers in version 0.1.8 with commit #2477, but usage instructions are not updated on their MCP server tutorial page yet, neither on here as well. 

I've tested and remote MCP server for context7 works with Gemini CLI.


enesgules (approved):
TYSM!

PR #401, opened 2025-07-13
Add dependabot config

yu-iskw:
It would be great to automaitcally upgrade outdated resources with dependabot.


enesgules (approved):
I think this is a good addition, I just hope it doesn't create too much PR noise

PR #413, opened 2025-07-16
fix: adding-projects link

shihyuho:
Fixes the links to the `adding-projects.md` document in the Japanese, Korean, and Traditional Chinese README files.


enesgules (approved):
TYSM

Issue #250, opened 2025-06-02
Proxy support


Coda1304:
Hello everyone,

at work we have to use a proxy to connect to the internet. I didn't see any option to set a proxy in the documentation. I also tried to set the environment variables `HTTP_PROXY` and `HTTPS_PROXY`, but they were not respected. From my tests I assume that there is no official proxy support and therefore have two questions:

1. Is there any official proxy support which I missed? Maybe even some documentation?
2. Could proxy support be added?

Thank you very much.

The error log which I get at the moment:

> 2025-06-02 15:31:03.576 [warning] [server stderr] Error searching libraries: TypeError: fetch failed
2025-06-02 15:31:03.576 [warning] [server stderr]     at node:internal/deps/undici/undici:13510:13
2025-06-02 15:31:03.576 [warning] [server stderr]     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
2025-06-02 15:31:03.576 [warning] [server stderr]     at async searchLibraries (file:///app/dist/lib/api.js:12:26)
2025-06-02 15:31:03.576 [warning] [server stderr]     at async file:///app/dist/index.js:55:28
2025-06-02 15:31:03.576 [warning] [server stderr]     at async file:///app/node_modules/@modelcontextprotocol/sdk/dist/esm/server/mcp.js:84:30 {
2025-06-02 15:31:03.576 [warning] [server stderr]   [cause]: Error: connect ECONNREFUSED 76.76.21.21:443
2025-06-02 15:31:03.576 [warning] [server stderr]       at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1636:16) {
2025-06-02 15:31:03.576 [warning] [server stderr]     errno: -111,
2025-06-02 15:31:03.576 [warning] [server stderr]     code: 'ECONNREFUSED',
2025-06-02 15:31:03.576 [warning] [server stderr]     syscall: 'connect',
2025-06-02 15:31:03.577 [warning] [server stderr]     address: '76.76.21.21',
2025-06-02 15:31:03.577 [warning] [server stderr]     port: 443
2025-06-02 15:31:03.577 [warning] [server stderr]   }
2025-06-02 15:31:03.577 [warning] [server stderr] }

enesgules:
Thanks for the recommendation. There is no official proxy support yet. We will add it as soon as we can

clemux:
[...]

PR #424, opened 2025-07-17
Honor https_proxy environment variables

sregister:
This should allow the mcp server to operate behind http proxy servers. If the standard https_proxy environment variable is defined, it should be used for requests to https://context7.com/api


enesgules:
Would this solve https://github.com/upstash/context7/issues/250 ?
Also could you resolve the conflicts please?

sregister:
Yes I think this should fix the problem described in #250

enesgules:
I guess the bun lockfile needs to be updates as well for the workflow to run successfully

sregister:
> I guess the bun lockfile needs to be updates as well for the workflow to run successfully

I think this should be fixed now

enesgules:
Could you update the readme to document this proxy support please, also please run the linter before your final commit

sregister:
I've added a note to README that standard https_proxy environment variables should be used.

tricktron:
@sregister @enesgules I resolved the conflict on my fork and can confirm that it works behind our corporate proxy.

enesgules:
Awesome @tricktron thanks for checking, will merge today

enesgules (approved):
TYSM!

PR #429, opened 2025-07-18
Update README.md Add details for Jetbrains Junie configuration

homich1991:
Add details for Jetbrains Junie configuration


enesgules (approved):
TYSM!

Issue #385, opened 2025-07-09
Context7 MCP gets stuck at connecting


n1lanjan:
# Steps to reproduce

- Go to https://playground.ai.cloudflare.com/ 
- Type https://mcp.context7.com/mcp and try to connect MCP server
- It gets stuck at Connecting. 

<img width="434" height="483" alt="Image" src="https://github.com/user-attachments/assets/9a13ecbb-b35f-4cd4-8f77-e7e38c284588" />

Its probably because of CORS Policy as I see this in the console error logs

- CORS policy: Request header field mcp-protocol-version is not allowed by Access-Control-Allow-Headers

enesgules:
it tries to connect via sse but the /mcp route is http, could you try with the /sse route please

n1lanjan:
The /sse route also fails likely because of the same CORS issues:

<img width="470" height="504" alt="Image" src="https://github.com/user-attachments/assets/59e45033-b906-49db-a205-e370c82a10d4" />

iamweird:
Specifying https://mcp.context7.com/sse URL solved the issue with Cline for me.

btw Context7 is not listed in the Cline MCP Server Marketplace as it is stated in the Context7 readme.

enesgules:
That's really weird, we used to be

enesgules:
<img width="1185" height="689" alt="Image" src="https://github.com/user-attachments/assets/da9cc249-31ae-44a4-b498-62231cd3d6f3" />

We are listed here

enesgules:
This is deployed btw, thanks for your patience

PR #440, opened 2025-07-21
Correct CORS headers required for connecting from browser based MCP Clients

n1lanjan:
As per 

- https://github.com/modelcontextprotocol/typescript-sdk/blob/main/README.md#cors-configuration-for-browser-based-clients

image

This change should fix the following

- https://github.com/upstash/context7/issues/385

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -213,7 +213,7 @@ async function main() {
       // Set CORS headers for all responses
       res.setHeader("Access-Control-Allow-Origin", "*");
       res.setHeader("Access-Control-Allow-Methods", "GET,POST,OPTIONS,DELETE");
-      res.setHeader("Access-Control-Allow-Headers", "Content-Type, mcp-session-id");
+      res.setHeader("Access-Control-Allow-Headers", "Content-Type, MCP-Session-Id, MCP-Protocol-Version");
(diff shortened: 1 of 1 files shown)


enesgules:
Hey @n1lanjan this is great, I didn't know we had to do that thanks for your contribution!

Do you think if we should update the OPTIONS responses to send the accepted headers as well? Also should we keep the uppercased header in the allowed headers just in case?

n1lanjan:
The OPTIONS responses would also send the accepted headers as per the current implementation because the check if the request is an OPTIONS request is after setting the headers.

HTTP headers are case-insensitive, so I think it should be fine to just keep one of the two, but no harm keeping both. I updated the PR to retain the existing headers.

enesgules:
Thanks that makes sense. Could you explain what is the MCP-Protocol-Version header as well please?

n1lanjan:
The latest [MCP Protocol version 2025-06-18 ](https://modelcontextprotocol.io/specification/2025-06-18/changelog) added Version Negotiation requiring the MCP-Protocol-Version header

- https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#version-negotiation

enesgules:
Thanks for all the explanation!

enesgules:
We will redeploy the remote server as soon as we can

PR #580, opened 2025-08-22
docs: update README.zh-CN.md

BiFangKNT:


enesgules:
Could you add the api key mentions as well please

BiFangKNT:
> Could you add the api key mentions as well please

Could you please specify which part you mean? I don't quite understand

enesgules:
In the main README we have CONTEXT7_API_KEY mentions all over, it would be cool if you could add those

BiFangKNT:
> In the main README we have CONTEXT7_API_KEY mentions all over, it would be cool if you could add those

May I ask what this CONTEXT7_API_KEY is used for? I noticed that some examples of "Remote Server Connection" do not include this parameter, while others do. Is this an oversight or by design?

enesgules:
This is an API key that helps you to configure your experience using context7. you can get it at context7.com/dashboard 

some clients don't support sending headers so we did not mention the api key for those examples but if you think there are missing configs please create a separate pr for the main readme

BiFangKNT:
You said:

> some clients don't support sending headers

But I noticed that some client examples include this parameter in the Local Server Connection section, while the Remote Server Connection section does not, for example:

> #### Roo Code Remote Server Connection
> 
> ```json
> {
>   "mcpServers": {
>     "context7": {
>       "type": "streamable-http",
>       "url": "https://mcp.context7.com/mcp"
>     }
>   }
> }
> ```
>
> #### Roo Code Local Server Connection
> 
> ```json
> {
>   "mcpServers": {
>     "context7": {
>       "command": "npx",
>       "args": ["-y", "@upstash/context7-mcp", "--api-key", "YOUR_API_KEY"]
>     }
>   }
> }
> ```

These are the same client, but one has parameters while the other does not. Is this an oversight or intentional design?

BiFangKNT:
Additionally, according to my understanding, this key is used to increase the request rate limit when using your cloud service, so why is it required to fill in this key for local deployment?

enesgules:
It's optional

enesgules:
Roo code doesn't support sending headers afaik through the remote server

enesgules:
This got way more complex than it needed to be :D

BiFangKNT:
> It's optional

Please forgive me for not fully understanding the operating principle of this project. If I am not mistaken, local deployment also requires calling your cloud service to retrieve data? If that's the case, it makes sense. At first, I thought that during local deployment, the related document collection and processing logic were local and did not need to send requests to your cloud service.

enesgules:
No worries at all!

sakuyamaij:
Line 43 need to be fix in README.zh-CN.md
`./docs/adding-projects.md` will redirect to the wrong link.
`./adding-projects.md` is correct

BiFangKNT:
> Line 43 need to be fix in README.zh-CN.md `./docs/adding-projects.md` will redirect to the wrong link. `./adding-projects.md` is correct

fixed

enesgules (approved):
Thank you so much! Especially for all the rafactors!

PR #703, opened 2025-09-21
Update README.md

Kurry:
fix Kiro documentation, the original causes the issue below:

```
2025-09-21 10:42:28.932 [info] [Context7] Reconnecting to server
2025-09-21 10:42:28.936 [info] [Context7] Registering MCP server and starting connection
2025-09-21 10:42:30.428 [warning] [Context7] Log from MCP Server: error: too many arguments. Expected 0 arguments but got 2.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,8 +496,13 @@ See [Kiro Model Context Protocol Documentation](https://kiro.dev/docs/mcp/config
   "mcpServers": {
     "Context7": {
       "command": "npx",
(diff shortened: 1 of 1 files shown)


enesgules:
Thanks for the report and the fix. just to double check, could you check the version of the mcp server? you can add a @latest tag at the end as well. I'm suspecting that it might be an older version maybe?

enesgules:
Hey @Kurry any updates?

PR #723, opened 2025-09-29
Add MacOS configuration notes for timeout issues

farion1231:
## Problem
I encountered request timeout errors on macOS when using Context7 MCP with Claude Desktop (Codex CLI). The issue occurs when Context7 needs to be re-downloaded due to version changes or updates, and the download time exceeds Codex's default timeout settings.

## Root Cause
- When Context7 package version changes or needs to be reinstalled
- The download/installation process takes longer than Codex's default timeout limit
- This results in "request timed out" errors, preventing the MCP server from initializing properly

## Solution
Similar to the existing Windows solution, the fix involves specifying the full paths to both Node.js and the installed Context7 package in the configuration. This bypasses the initialization timeout by directly pointing to the already-installed resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -493,9 +493,27 @@ startup_timeout_ms = 20_000
 
 This ensures Codex CLI works reliably on Windows.
 
+⚠️ MacOS Notes
+
+On MacOS, some users may encounter the same request timed out errors like Windows,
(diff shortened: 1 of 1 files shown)


enesgules (approved):
Thanks for the update!

PR #760, opened 2025-10-09
docs: add Neovate Code installation instructions

xierenyuan:
Adding instructions for how to setup the MCP server for [Neovate Code](https://neovateai.dev/en/docs/overview/)


enesgules:
Thanks for the contribution but we don't want to overpopulate the readme

PR #776, opened 2025-10-15
feat: support gemini-extension file

enesgules:
fixes ctx7-661

### DIFF
--- a/gemini-extension.json
+++ b/gemini-extension.json
@@ -9,7 +9,7 @@
           "-y",
           "@upstash/context7-mcp",
           "--api-key",
-          "YOUR_API_KEY"
+          "${CONTEXT7_API_KEY}"
         ]
       }
     }
--- a/gemini-extension.json
+++ b/gemini-extension.json
@@ -1,0 +1,16 @@
+{
+    "name": "context7",
+    "description": "Up-to-date code docs for any prompt",
+    "version": "1.0.0",
+    "mcpServers": {
+      "context7": {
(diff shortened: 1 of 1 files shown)


gemini-extension.json
```diff
@@ -0,0 +1,16 @@
...
+    "description": "Up-to-date code docs for any prompt",
+    "version": "1.0.0",
+    "mcpServers": {
+      "context7": {
+        "command": "npx",
+        "args": [
+          "-y",
+          "@upstash/context7-mcp",
+          "--api-key",
+          "YOUR_API_KEY"
```
chrstnb: I would recommend having the API key stored in a named environment variable and then instructing the user to set it in your README

enesgules: Hi @chrstnb, thanks for the review!
We already support an env variable called CONTEXT7_API_KEY
If a user has this in their env, would the mcp server be able to use it through the extension? If so, should I just remove this flag?

chrstnb: I think the common pattern is to pass it in via the MCP definition (so like you're doing here but just with the CONTEXT7_API_KEY)

enesgules: I couldn't understand exactly, would the following work?
```
"--api-key",
"${CONTEXT7_API_KEY}"
```

If not could you share the config that you meant?

chrstnb: Yes it should! And then just include in the README that that env var needs to be set

chrstnb (approved):

PR #849, opened 2025-10-26
feat: add remote server connection for OpenAI Codex across all docs

someblue:
Add a remote server configuration example for `Codex` so that users who fail to execute npx commands in their local environment can access it via HTTP.


enesgules (approved):
TYSM!

PR #874, opened 2025-10-30
make call to action more visible

omerXfaruq:
call to action buttons are not very visible between all other banners, move them to top

before:

image

now:

image


enesgules (approved):
Thank you so much!

PR #926, opened 2025-11-06
Add warnings against mistaking context7 as part of the question

omerXfaruq:
Previously llm was mistaking context7 as part of the question for prompts like this: `use context7 TypeScript Node.js MCP general request failed error`
This PR adds warnings against this situation in the description and response of resolve-library-id. After the changes llm doesn't mistake context7 as part of the question anymore. Happy to share the output if needed.

Update readme for local mcp setup as the previous instructions were not working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1301,7 +1301,8 @@ CONTEXT7_API_KEY=your_api_key_here
   "mcpServers": {
     "context7": {
       "command": "npx",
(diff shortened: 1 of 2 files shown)


omerXfaruq:
cc: @enesakar @enesgules

enesgules:
Closing as the other pr contiains the same contents too #935

PR #935, opened 2025-11-06
Update readme with specific package version usage example

omerXfaruq:
Improve readme.

Working prompt example:

image

There is only one problem, for each package, the number of versions available on context7 is very limited. If that version doesn't exist in context7 it'll default to no version.


omerXfaruq:
cc: @enesakar, @enesgules

enesakar:
Thanks for the contribution. We are handling this in another, more general prompt improvement development, so closing here.

PR #1047, opened 2025-11-19
docs: Update Antigravity server configuration to correct struct

ovftank:


enesgules:
I think the one you added is the vs code default mcp config which doesn't work with antigravity. Did you test this config? Were you able to use context7?

ovftank:
> I think the one you added is the vs code default mcp config which doesn't work with antigravity. Did you test this config? Were you able to use context7?

i use `workbench.mcp.openUserMcpJson` options and it working(`antigravity.openMcpConfigurationPanel` blank when choose, maybe bug from antigravity), in latest version Antigravity and in clean windows device
image

enesgules:
here is my twitter post displaying both configs, unfortunately the ide is a bit broken https://x.com/abdushbag/status/1991110226200986029?s=20

ovftank:
> here is my twitter post displaying both configs, unfortunately the ide is a bit broken https://x.com/abdushbag/status/1991110226200986029?s=20

in my case, it only working with default config from vscode ..., maybe it is bug from antigravity,so ... just wait them fix it!

PR #1061, opened 2025-11-22
CTX7-855: TypeScript SDK

fahreddinozcan:


packages/sdk/src/http/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    const signal = this.options.signal;
+    const isSignalFunction = typeof signal === "function";
+
+    const requestOptions = {
+      cache: this.options.cache,
+      method: "POST",
+      headers: this.headers,
+      body: JSON.stringify(req.body),
+      keepalive: true,
+      signal: isSignalFunction ? signal() : signal,
```
CahidArda: I had added the isSignalFunction for backwards compatibility to our other SDKs

It should always be a function, you can remove non-function `Signal` from `HttpClientConfig["signal"]`

Reason is: If you try using with a non function and get a timeout, signal will always stay as `aborted: true` in subsequent requests. Passing a function allows us to create a fresh signal for every request

packages/sdk/src/http/index.ts
```diff
@@ -0,0 +1,156 @@
...
+          };
+  }
+
+  public async request<TResult>(req: Context7Request): Promise<Context7Response<TResult>> {
+    const signal = this.options.signal;
+    const isSignalFunction = typeof signal === "function";
+
+    const requestOptions = {
+      cache: this.options.cache,
+      method: "POST",
```
CahidArda: some of our endpoints are GET I think

packages/sdk/src/types.ts
```diff
@@ -0,0 +1,66 @@
...
+  baseUrl?: string;
+  timeout?: number;
+  encryptionKey?: string;
+  clientIp?: string;
+}
+
+export interface SearchLibraryOptions {
+  limit?: number;
+  apiKey?: string;
+  clientIp?: string;
```
CahidArda: these two probably shouldn't be part of options, we don't use them too

packages/sdk/src/commands/get-docs/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    if (parts.length < 2) {
+      throw new Error(
+        `Invalid library ID format: ${libraryId}. Expected format: /username/library or /username/library/version`
+      );
+    }
+
+    const owner = parts[0];
+    const repo = parts[1];
+    const version = parts[2] || options?.version;
```
CahidArda: to keep things simple, we could only allow `/username/library`, check for `parts.length !== 2`, and set `version` to `options.version`

packages/sdk/src/commands/types.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export interface GetDocsOptions {
+  version?: string;
+  page?: number;
+  topic?: string;
+  limit?: number;
+  docType?: "info" | "code";
+  format?: "json" | "txt";
+}
```
CahidArda, line 104: would be nice to document:
- the defaults, for docType and others if they have one
- that version will be ignored if you pass /username/library/version`

packages/sdk/src/commands/get-docs/index.ts
```diff
@@ -0,0 +1,53 @@
...
+      queryParams.limit = options.limit;
+    }
+
+    super(
+      {
+        method: "GET",
+        path: pathParts,
+        query: queryParams,
+      },
+      pathParts.slice(0, 3).join("/")
```
CahidArda: edit: got it, this is not the endpoint we send a request to, that's 3 lines above. But why is there a second endpoint way to pass endpoint in the first place? I would just allow path above

pathParts starts as `const pathParts = ["v2", "docs", docType, owner, repo];`, and we add version. This slice looks wrong to me

Not sure how the tests passed if this is wrong

packages/sdk/src/http/index.ts
```diff
@@ -0,0 +1,177 @@
...
+   */
+  method?: "GET" | "POST";
+  /**
+   * Query parameters for GET requests
+   */
+  query?: Record<string, string | number | boolean | undefined>;
+};
+export type Context7Response<TResult> = {
+  result?: TResult;
+  error?: string;
```
CahidArda: there are no references of this field, it can be removed

alternatively, you can return `{ error }` in request instead of throwing (like [here](https://github.com/upstash/context7/blob/cb63f0da8bf1e502e4a000dd9ca88c3005e8f9b5/packages/sdk/src/http/index.ts#L150)). Then you can handle the error and undefined result together [here](https://github.com/upstash/context7/blob/cb63f0da8bf1e502e4a000dd9ca88c3005e8f9b5/packages/sdk/src/commands/command.ts#L35).

I would go with just deleting error :D

packages/sdk/src/client.ts
```diff
@@ -0,0 +1,71 @@
...
+    libraryId: string,
+    options: GetDocsOptions & { format: "json"; docType: "info" }
+  ): Promise<InfoSnippetsResponse>;
+
+  async getDocs(
+    libraryId: string,
+    options: GetDocsOptions & { format: "json"; docType?: "code" }
+  ): Promise<CodeSnippetsResponse>;
+
+  async getDocs(libraryId: string, options?: GetDocsOptions & { format?: "txt" }): Promise<string>;
```
CahidArda: in the response, can't there be pagination for txt format too?

fahreddinozcan: Currently, the endpoint returns just a txt for txt format query. So there's no pagination metadata in the response. we may change this in the future.

packages/sdk/src/http/index.ts
fahreddinozcan: removed the error 👍🏻

packages/sdk/src/commands/types.ts
fahreddinozcan: done 👍🏻

packages/sdk/src/client.ts
CahidArda: Then this could become a breaking change in the future

I think it makes sense to return in a format similar to the other two (without pagination) to avoid a breaking change

fahreddinozcan: I will add a paging support for txt, by using headers. so the headers on response will look like
```
headers: { 
    "Content-Type": "text/plain; charset=utf-8",
    "X-Page": String(result.data.pagination.page),
    "X-Limit": String(result.data.pagination.limit),
    "X-Total-Pages": String(result.data.pagination.totalPages),
    "X-Has-Next": String(result.data.pagination.hasNext),
    "X-Has-Prev": String(result.data.pagination.hasPrev),
  }
```

we'll also use the same headers on reqs too.

CahidArda: User can't access these headers/fields since we just return `Promise<string>`. If we return `Promise<string>` now, this will still be a breaking change unless I am missing something. `Promise<string>` will have to turn into something like

```typescript
type TextResponse = {
  text: string;
  totalTokens: number;
  pagination: Pagination;
  metadata: APIResponseMetadata;
}
```

or we will have to add a new parameter (that we didn't need in other formats) to include the response above or `Promise<string>`

We can return the type I wrote above without pagination, and add pagination later if needed without a breaking change

packages/sdk/src/client.ts
```diff
@@ -0,0 +1,75 @@
...
+  ): Promise<InfoDocsResponse>;
+
+  async getDocs(
+    libraryId: string,
+    options: GetDocsOptions & { format: "json"; docType?: "code" }
+  ): Promise<CodeDocsResponse>;
+
+  async getDocs(
+    libraryId: string,
+    options?: GetDocsOptions & { format?: "txt" }
```
CahidArda: if docType has no effect when format is txt, we can dissallow it with
```suggestion
    options?: GetDocsOptions & { format?: "txt"; docType?: undefined }
```

CahidArda (approved):

enesakar (approved):

PR #1071, opened 2025-11-24
refactor: centralize error message generation and simplify code

enesgules:
fixes ctx7-863

### DIFF
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -2,6 +2,7 @@ import { SearchResponse } from "./types.js";
 import { generateHeaders } from "./encryption.js";
 import { ProxyAgent, setGlobalDispatcher } from "undici";
 import { DocumentationMode, DOCUMENTATION_MODES } from "./types.js";
+import { maskApiKey } from "./utils.js";
 
 const CONTEXT7_API_BASE_URL = "https://context7.com/api";
 const DEFAULT_TYPE = "txt";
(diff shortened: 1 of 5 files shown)


src/lib/api.ts
```diff
@@ -34,6 +33,35 @@ function parseLibraryId(libraryId: string): {
...
+): string {
+  switch (errorCode) {
+    case 429:
+      return apiKey
+        ? "Rate limited due to too many requests. Please try again later."
+        : "Rate limited due to too many requests. You can create a free API key at https://context7.com/dashboard for higher rate limits.";
+    case 404:
+      return "The library you are trying to access does not exist. Please try with a different library ID.";
+    case 401:
+      return `Unauthorized. Please check your API key. The API key you provided (possibly incorrect) is: ${apiKey}. API keys should start with 'ctx7sk'`;
```
fahreddinozcan:
```suggestion
      return `Unauthorized. Please check your API key. API keys should start with 'ctx7sk'`;
```
let's not include the user's token in the error message, they might render this error somewhere for their users etc.

src/lib/api.ts
```diff
@@ -34,6 +33,35 @@ function parseLibraryId(libraryId: string): {
...
+    default:
+      if (docMode) {
+        return `Failed to fetch ${docMode} documentation. Please try again later. Error code: ${errorCode}`;
+      }
+      return `Failed to fetch documentation. Please try again later. Error code: ${errorCode}`;
```
fahreddinozcan, line 57:
```suggestion
    default:
      return `Failed to fetch documentation. Please try again later. Error code: ${errorCode}`;
```
imo this would be enough

fahreddinozcan (approved):

PR #1104, opened 2025-11-28
docs: code rabbit integration

enesgules:


docs/docs.json
```diff
@@ -32,6 +32,10 @@
           "dashboard/billing"
         ]
       },
+      {
+        "group": "Integrations",
+        "pages": ["integrations/code-rabbit"]
+      },
       {
         "group": "API Reference",
```
fahreddinozcan: Should we put integrations after the API Reference?

enesgules: makes sense

fahreddinozcan (approved):

PR #1176, opened 2025-12-05
CTX7-884: OAuth Support for MCP

fahreddinozcan:
This PR adds OAuth 2.0 authentication support to the MCP server, enabling secure access control for MCP clients.

### Changes
**New OAuth-protected endpoint (/mcp/oauth)** - Requires authentication via JWT tokens
**JWT validation** - Validates tokens against the authorization server's JWKS endpoint with caching (1 hour TTL)
**OAuth Protected Resource Metadata** - Implements RFC 9728 discovery endpoint at /.well-known/oauth-protected-resource
**WWW-Authenticate header** - Returns resource metadata URL to help clients discover the authorization server
**Backward compatibility** - The existing /mcp endpoint remains available for anonymous access

### New Files
`packages/mcp/src/lib/jwt.ts` - JWT validation utilities using jose library

### Configuration
`AUTH_SERVER_URL` - Authorization server URL (default: https://context7.com)
`RESOURCE_URL` - Resource server URL for metadata


packages/mcp/src/lib/api.ts
```diff
@@ -2,7 +2,8 @@ import { SearchResponse, ContextRequest, ContextResponse } from "./types.js";
 import { ClientContext, generateHeaders } from "./encryption.js";
 import { ProxyAgent, setGlobalDispatcher } from "undici";
 
-const CONTEXT7_API_BASE_URL = "https://context7.com/api";
+// const CONTEXT7_API_BASE_URL = process.env.CONTEXT7_API_URL || "https://context7.com/api";
+const CONTEXT7_API_BASE_URL = "http://localhost:3000/api";
```
enesgules: lets fix this

packages/mcp/src/lib/jwt.ts
```diff
@@ -0,0 +1,66 @@
+import * as jose from "jose";
+
+// Clerk is the OAuth token issuer
+const CLERK_DOMAIN = process.env.CLERK_DOMAIN || "clerk.context7.com";
```
enesgules: let's hardcode this instead of env var

packages/mcp/src/index.ts
```diff
@@ -316,11 +317,52 @@ async function main() {
...
 
-    app.all("/mcp", async (req: express.Request, res: express.Response) => {
+    const handleMcpRequest = async (
+      req: express.Request,
+      res: express.Response,
+      requireAuth: boolean
+    ) => {
       try {
+        const apiKey = extractApiKey(req);
+        const resourceUrl = process.env.RESOURCE_URL || `http://localhost:${initialPort}`;
```
enesgules: should we use env var here or hardcode it?

packages/mcp/src/index.ts
```diff
@@ -348,12 +390,39 @@ async function main() {
...
       res.json({ status: "ok", message: "pong" });
     });
 
+    // OAuth 2.0 Protected Resource Metadata (RFC 9728)
+    // Used by MCP clients to discover the authorization server
+    app.get(
+      "/.well-known/oauth-protected-resource",
+      (_req: express.Request, res: express.Response) => {
+        const authServerUrl = process.env.AUTH_SERVER_URL || "https://context7.com";
+        const resourceUrl = process.env.RESOURCE_URL || "https://mcp.context7.com";
```
enesgules: do these should be env vars?

packages/mcp/src/lib/jwt.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export function isJWT(token: string): boolean {
+  const parts = token.split(".");
+  return parts.length === 3;
+}
+
+export async function validateJWT(token: string): Promise<JWTValidationResult> {
+  if (!isJWT(token)) {
+    return { valid: false, error: "Not a valid JWT format" };
+  }
```
enesgules: is this call redundant since we call validateJWT after calling isJWT?

packages/mcp/src/lib/jwt.ts
```diff
@@ -0,0 +1,66 @@
...
+  jwksCache = jose.createRemoteJWKSet(new URL(JWKS_URL));
+  jwksCacheTime = now;
+  return jwksCache;
+}
+
+export interface JWTValidationResult {
+  valid: boolean;
+  userId?: string;
+  clientId?: string;
+  scope?: string;
```
enesgules: userId, clientId, scope are extracted but never used

packages/mcp/src/lib/jwt.ts
```diff
@@ -0,0 +1,66 @@
...
+async function getJWKS(): Promise<jose.JWTVerifyGetKey> {
+  const now = Date.now();
+  if (jwksCache && now - jwksCacheTime < JWKS_CACHE_TTL) {
+    return jwksCache;
+  }
+
+  jwksCache = jose.createRemoteJWKSet(new URL(JWKS_URL));
+  jwksCacheTime = now;
+  return jwksCache;
+}
```
enesgules: Why do we need this cache? Claude said jose's default (30s cooldown + auto-refresh on unknown kid) is generally better for security since it handles key rotation faster. Also since our MCP server is stateless, different requests go to different servers so not sure if we need this 1 hour cache.
Also the 1-hour custom cache is actually caching the function (createRemoteJWKSet), not the JWKS data itself. Jose still caches internally within that function. So you have double caching that doesn't add much value.

enesgules (approved):

PR #1183, opened 2025-12-08
Python SDK

fahreddinozcan:
create python sdk


packages/python/sdk/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+async def main():
+    async with Context7(api_key="ctx7sk_...") as client:
+        # Search for libraries
+        results = await client.search_library("react")
+        for lib in results.results:
+            print(f"{lib.id}: {lib.title}")
+
+        # Get documentation
+        docs = await client.get_docs("/facebook/react")
```
enesgules: should this be get_context?

.github/workflows/changeset-check.yml
```diff
@@ -34,8 +34,20 @@ jobs:
               { encoding: 'utf-8' }
             );
 
-            const hasChangeset = diff
-              .split('\n')
+            const changedFiles = diff.split('\n').filter(f => f.length > 0);
+
+            // Skip changeset check if only Python files or workflow files are changed
```
enesgules: maybe we can skip changeset if only markdown files are updated

packages/python/sdk/README.md
```diff
@@ -0,0 +1,69 @@
...
+### `client.search_library(query: str, library_name: str) -> list[Library]`
+
+Search for libraries matching the given query.
+
+- `query`: The user's question or task (used for relevance ranking)
+- `library_name`: The library name to search for
+
+Async version: `await client.search_library_async(query, library_name)`
+
+### `client.get_context(query: str, library_id: str, *, type="txt") -> str | list[Documentation]`
```
enesgules, line 55: default is json right?

packages/python/sdk/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Get documentation context for a library.
+
+**Parameters:**
+
+- `query`: The user's question or task
+- `library_id`: Context7 library ID (e.g., `/facebook/react`)
+- `type`: Response format - `"txt"` (default) returns plain text, `"json"` returns list of Documentation objects
```
enesgules, line 63: also here

enesgules (commented):
The TS SDK has type option for searchLibrary and Python doesn't do we need to add it to here too? What do you think?

enesgules (commented):
Python SDK has a configurable base_url but TS SDK doesn't have it. Should we remove this or add it to TS SDK too?

enesgules (approved):

PR #1319, opened 2025-12-29
docs: update tool name to query-docs in install guides

yeepchen:
Following the official renaming of the tool get-library-docs to query-docs, this PR updates all references within the context7 project documentation to maintain consistency and ensure the installation instructions are accurate for users.


enesgules:
Hey thanks for the contribution but we have removed the page to redirect to our readme so no need for an update on this front. Thanks!

PR #1500, opened 2026-01-19
Clarify server types and update installation instructions

omerXfaruq:
While installing context7 in claude code i was confused and installed the local version by mistake.

Renamed server types for clarity and reordered sections to prioritize the recommended method. Updated installation instructions for various platforms.


omerXfaruq:
cc: @enesakar @enesgules

enesgules:
Hi, thanks for the contribution but there a few issues on the pr. 
1. API key doesn't guarantee persistent context
2. We don't want to offer the npm package as an advanced feature, it's still quite simple to use for many people
3. OAuth is also available for regular users, not just teams/enterprises

That's why closing this. Thanks again and Ramadan Mubarak!

PR #1523, opened 2026-01-22
feat: add installation count metric collection

fahreddinozcan:


packages/cli/src/utils/api.ts
```diff
@@ -30,6 +30,15 @@ export async function searchSkills(query: string): Promise<SearchResponse> {
   return (await response.json()) as SearchResponse;
 }
 
+export function trackInstalls(skills: string[], ides: string[]): void {
+  if (process.env.CTX7_TELEMETRY_DISABLED || !skills.length) return;
```
enesgules, line 34: maybe we can rename this variable to CONTEXT7_TELEMETRY_DISABLED

enesgules (approved):

packages/cli/src/utils/api.ts
fahreddinozcan: cli is CTX7, so i think this is the proper naming. people can understand this is related to cli

PR #1596, opened 2026-01-29
add tool to compare two libraries

fernandaaraujo0202:
# Descrição

I've added a new feature to the MCP server that allows users to compare two libraries directly. This makes it easier to evaluate different tools using Context7 data.

## Main change:

* **New tool:** I implemented a handler in src/index.ts that accepts libraryIdA and libraryIdB;
     * The tool performs searches in both libraries in parallel;
     
* **Type safety:**  Updated src/index.ts and src/lib/utils.ts with the necessary interfaces and types to support the new feature and ensure the project compiles without errors.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -5,7 +5,7 @@ import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js"
 import { z } from "zod";
 import { searchLibraries, fetchLibraryContext } from "./lib/api.js";
 import { ClientContext } from "./lib/encryption.js";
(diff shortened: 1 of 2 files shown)


fahreddinozcan:
Hey @fernandaaraujo0202, 
I think MCP can currently do this comparison, because searchLibrary tool returns multiple libraries at once, so they're comparable in single prompt.  I'm curious why you think this tool is necessary?

we try to keep the MCP tools as concise and less as possible to make sure we don't bloat your context. So we are not adding new tools unless its really necessary.

PR #1618, opened 2026-01-30
feat(cli): improve skill search display and revamp generate UX

fahreddinozcan:
## Summary

- **Trust score & installs columns**: Add Installs and Trust(0-10) columns to skill search results with aligned column headers
- **Auto-login for generate**: Instead of showing an error when unauthenticated, automatically trigger the OAuth login flow
- **Recommended option first**: Reorder question options so the recommended choice is always at the top
- **View in editor**: Open generated skill content in the user's default editor (`$EDITOR` or `open` fallback) instead of dumping inline
- **Generate UX revamp**: Improved copy, do/don't examples, renamed "libraries" to "sources", clearer spinner text

### DIFF
--- a/packages/cli/src/types.ts
+++ b/packages/cli/src/types.ts
@@ -8,12 +8,11 @@ export interface Skill {
   description: string;
   url: string;
   installCount?: number;
+  trustScore?: number;
 }
 
 export interface SkillSearchResult extends Skill {
   project: string;
-  installCount?: number;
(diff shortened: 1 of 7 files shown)


packages/cli/src/types.ts
enesgules, line 6: maybe we can remove the install count from SkillSearchResult and move trust score to skill object too

enesgules (approved):

PR #1667, opened 2026-02-03
feat(cli): use backend suggest endpoint

fahreddinozcan:
## Summary
- Replace client-side NOTABLE_SET/SKIP_SET with single API call to `/api/v2/skills/suggest`
- Backend now handles filtering, vector search, and ranking
- Simplify `deps.ts` to only collect dependencies without filtering

### DIFF
--- a/packages/cli/src/types.ts
+++ b/packages/cli/src/types.ts
@@ -194,13 +194,12 @@ export const DEFAULT_CONFIG: C7Config = {
 };
 
 // Suggest endpoint types
-export interface SuggestGroup {
-  dependency: string;
-  skills: SkillSearchResult[];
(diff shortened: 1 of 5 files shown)


packages/cli/src/types.ts
```diff
@@ -142,6 +142,7 @@ export interface ScopeOptions {
 }
 
 export type AddOptions = IDEOptions & ScopeOptions & { all?: boolean };
+export type DiscoverOptions = IDEOptions & ScopeOptions;
```
enesgules: i guess this is an older name, maybe we can call it suggest options? but it's your call

enesgules (approved):

PR #1684, opened 2026-02-03
Remove deprecated downlevelIteration option

RyanCavanaugh:
This option is deprecated in TypeScript 6.0. It doesn't do anything when `target` is `esnext`, so can be safely removed from this config file.


enesgules:
Thank you so much!

enesgules (approved):

PR #1741, opened 2026-02-07
HTTP transport: explicit JSON body limit + optional CORS allowlist

TheodorNEngoy:
Changes:
- Set an explicit Express JSON body limit for the HTTP transport (`express.json({ limit: "1mb" })`).
- Add optional CORS allowlisting via `CONTEXT7_ALLOWED_ORIGINS` (comma-separated). If unset, CORS stays permissive for compatibility.
- Document `CONTEXT7_ALLOWED_ORIGINS` in the developer guide.

Rationale: make the HTTP mode safer-by-default for self-hosting and reduce surprises around request size.

### DIFF
--- a/docs/resources/developer.mdx
+++ b/docs/resources/developer.mdx
@@ -63,6 +63,16 @@ You can use the `CONTEXT7_API_KEY` environment variable instead of passing the `
 - Integration with MCP server setups that use dotenv
 - Tools that prefer environment variable configuration
 
+### HTTP Transport CORS (Optional)
+
+When running with `--transport http`, you can optionally restrict browser origins by setting:
+
+```bash
+CONTEXT7_ALLOWED_ORIGINS="https://chatgpt.com,https://chat.openai.com"
+```
+
+If `CONTEXT7_ALLOWED_ORIGINS` is not set, CORS remains permissive for compatibility.
+
 <Warning>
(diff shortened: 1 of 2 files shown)


fahreddinozcan:
Hey Theodor,

can you please further explain the necessity and motivation of this PR? We don't return random data for response to MCP, so I can guarantee that response limit of 1 mb wouldn't be a case.

Do you have any specific requirement that would need cors? If you self host the mcp server, it's already on your local so no need for cors? If you want to expose your MCP server deployment, you can just use our deployment at https://mcp.context7.com/mcp. 

Anything I miss?

TheodorNEngoy:
Good questions, thanks.

1) The `1mb` limit is for the *incoming request JSON body* (Express’ JSON parser reads the full body into memory). It’s a basic DoS/footgun guard against accidentally accepting arbitrarily large payloads; it does not cap MCP responses.

2) CORS is only relevant for *browser* clients. Even for self-hosting, people sometimes run the HTTP transport on a LAN/remote box or access it from a web UI; permissive `Access-Control-Allow-Origin: *` means any website can read responses if the server is reachable. The allowlist is opt-in via `CONTEXT7_ALLOWED_ORIGINS` and keeps today’s permissive behavior if unset, so existing setups shouldn’t break.

If you’d rather keep the transport minimal, I’m happy to split this into two PRs (body limit only, or drop CORS entirely) based on your preference.

PR #1814, opened 2026-02-13
Add API key settings to gemini-extension.json

jarrodcolburn:
Added settings for API key configuration.


jarrodcolburn:
was missing `settings` section to set API key. see
https://geminicli.com/docs/extensions/reference/#settings

jarrodcolburn:
after this inclusion should be able to run `gemini extensions config` to be prompted to include API key

fahreddinozcan (approved):
Thank you!

PR #2025, opened 2026-02-26
Fix command syntax in README for local server connection

hendra-samudera:
Remove space before npx on claude code local server instalation


enesgules:
the one on the readme is the correct syntax per docs. https://code.claude.com/docs/en/mcp#option-3-add-a-local-stdio-server

hendra-samudera:
Oh ok my bad. Thanks for the reply 👍

PR #2056, opened 2026-03-01
fix: remove hardcoded encryption key, hash IPs instead of leaking plaintext

ashsolei:
## Summary

Removes hardcoded default encryption key and replaces plaintext IP fallbacks with SHA-256 hashing.

### Problem

The `DEFAULT_ENCRYPTION_KEY` constant (`000102...1e1f`) is checked into source code. Any deployment that doesn't explicitly set `CLIENT_IP_ENCRYPTION_KEY` silently uses this well-known key, rendering all client IP encryption meaningless — anyone with access to the source can decrypt the IPs.

Additionally, on any encryption failure (invalid key format, runtime error), the code falls back to returning the raw client IP in plaintext.

### Fix

1. **Removed `DEFAULT_ENCRYPTION_KEY`** — `ENCRYPTION_KEY` is now `undefined` when the env var is not set
2. **Added `hashClientIp()`** — Uses SHA-256 to one-way hash client IPs when encryption is unavailable
3. **All fallback paths now hash instead of returning plaintext** — covers missing key, invalid key format, and runtime encryption errors

### Behavior Change

| Scenario | Before | After |
|----------|--------|-------|
| No env var set | Encrypts with well-known key | SHA-256 hash (prefix: `sha256:`) |
| Invalid key format | Returns raw IP | SHA-256 hash |
| Encryption error | Returns raw IP | SHA-256 hash |
| Valid key set | AES-256-CBC encryption | AES-256-CBC encryption (unchanged) |

The SHA-256 hash preserves the ability to correlate requests from the same client IP while ensuring the IP address itself cannot be recovered.

### DIFF
--- a/packages/mcp/src/lib/encryption.ts
+++ b/packages/mcp/src/lib/encryption.ts
@@ -1,8 +1,7 @@
-import { createCipheriv, randomBytes } from "crypto";
+import { createCipheriv, randomBytes, createHash } from "crypto";
 import { SERVER_VERSION } from "./constants.js";
 
-const DEFAULT_ENCRYPTION_KEY = "000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f";
-const ENCRYPTION_KEY = process.env.CLIENT_IP_ENCRYPTION_KEY || DEFAULT_ENCRYPTION_KEY;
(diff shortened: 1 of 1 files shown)


ashsolei:

PR #2059, opened 2026-03-01
fix: create per-request McpServer in HTTP mode to prevent transport races

ashsolei:
The module-level `McpServer` singleton is shared across all concurrent HTTP requests. Each request calls `server.connect(transport)` which overwrites the server's internal `_transport` binding. When requests overlap, response A can be sent through request B's transport, delivering data to the wrong client.

## Root Cause

`McpServer.connect()` (from `@modelcontextprotocol/sdk`) binds the server to a single transport. With one shared server and per-request transports, the last `connect()` call wins — earlier requests lose their transport binding.

## Fix

Extract server creation and tool registration into a `createMcpServer()` factory function. In HTTP mode, each request now creates its own `McpServer` instance with an isolated transport binding.

- **Stdio mode**: Unchanged — uses the module-level singleton (only one transport ever exists)
- **HTTP mode**: Each request creates a fresh `McpServer` via `createMcpServer()`, connects its own transport, and handles the request in isolation

Since Context7 tools are stateless (they proxy to the Context7 API), there is no shared state concern with per-request server instances.

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -119,33 +119,28 @@ function getClientIp(req: express.Request): string | undefined {
   return undefined;
 }
 
-const server = new McpServer(
-  {
-    name: "Context7",
-    version: SERVER_VERSION,
-  },
-  {
-    instructions:
-      "Use this server to retrieve up-to-date documentation and code examples for any library.",
-  }
-);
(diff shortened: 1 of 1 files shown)


ashsolei:

PR #2102, opened 2026-03-04
fix(sdk): avoid raw SyntaxError on non-JSON error responses

### LINKED ISSUE #1964: Bug: SDK HttpClient throws raw SyntaxError instead of Context7Error on non-JSON error responses
## Bug Description

The SDK's `HttpClient.request()` method calls `res.json()` without a try/catch in the error-response path. When the server returns a non-JSON error body (HTML 502 gateway page, plain-text 429, Cloudflare challenge page, etc.), `res.json()` throws a native `SyntaxError` that bypasses the `Context7Error` type system.

## Affected Code

**File:** `packages/sdk/src/http/index.ts`, lines ~171-173

```typescript
if (!res.ok) {
  const errorBody = (await res.json()) as { error?: string; message?: string };
  throw new Context7Error(errorBody.error || errorBody.message || res.statusText);
}
```

`res.json()` is called bare — no try/catch, no Content-Type guard.

## Impact

- Any caller using `catch (error) { if (error instanceof Context7Error) ... }` will miss the error entirely
- Infrastructure errors (Cloudflare 502, AWS ALB 503, rate limiting 429 with HTML body) surface as confusing `SyntaxError: Unexpected token '<' at position 0` instead of a typed error with the HTTP status
- This affects all SDK consumers: the CLI, the MCP server, and any third-party integration using the SDK

## Steps to Reproduce

1. Configure Context7 SDK to point to a server that returns a non-JSON error
2. (Or simulate: proxy returning HTML 502 on the API endpoint)
3. Call any SDK method (e.g., `resolveLibraryId`)
4. Observe: `SyntaxError: Unexpected token '<'` instead of `Context7Error`

## Proposed Solution

Wrap `res.json()` in a try/catch:

```typescript
if (!res.ok) {
  const errorBody = await res.json().catch(() => ({})) as { error?: string; message?: string };
  throw new Context7Error(errorBody.error || errorBody.message || res.statusText);
}
```

This is a one-line change (adding `.catch(() => ({}))`) that ensures non-JSON error bodies gracefully fall through to the `res.statusText` fallback message, and the error is always a typed `Context7Error`.

Happy to submit a PR for this fix.

mitre88:
## Summary
- wrap SDK error JSON parsing in a safe try/catch path inside HttpClient.request()
- ensure non-JSON error bodies still produce a typed Context7Error
- add unit tests for JSON and non-JSON error response behavior

## Why
When the API returns HTML/plain-text error content, res.json() throws a raw SyntaxError. This bypasses Context7Error handling in consumers.

## Verification
- pnpm --filter @upstash/context7-sdk lint
- pnpm --filter @upstash/context7-sdk exec vitest run src/http/index.test.ts

Closes #1964

### DIFF
--- a/packages/sdk/src/http/index.test.ts
+++ b/packages/sdk/src/http/index.test.ts
@@ -1,0 +1,51 @@
+import { afterEach, describe, expect, test, vi } from "vitest";
+import { HttpClient } from "./index";
+import { Context7Error } from "@error";
+
+describe("HttpClient error handling", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  test("throws Context7Error for non-JSON error responses", async () => {
+    vi.spyOn(globalThis, "fetch").mockResolvedValue(
+      new Response("<html>Bad gateway</html>", {
+        status: 502,
+        statusText: "Bad Gateway",
+        headers: { "content-type": "text/html" },
+      })
+    );
+
+    const client = new HttpClient({
+      baseUrl: "https://example.com/api",
+      retry: false,
+    });
+
+    try {
+      await client.request({ path: ["v2", "libs", "search"] });
+      throw new Error("Expected request to throw");
+    } catch (error) {
+      expect(error).toBeInstanceOf(Context7Error);
+      expect((error as Error).message).toBe("Bad Gateway");
+    }
+  });
+
+  test("prefers API error message when response body is JSON", async () => {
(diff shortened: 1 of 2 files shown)


mitre88:
Gentle ping — this fix prevents raw `SyntaxError` on non-JSON error responses in the SDK. Small, targeted change. Would appreciate a review when possible. Thanks!

PR #2160, opened 2026-03-09
feat: add searchFilterApplied field to LibrarySearchResponse and SearchResponse interfaces; add warnings to cli and mcp

enesgules:


packages/cli/src/commands/docs.ts
```diff
@@ -91,6 +91,14 @@ async function resolveCommand(
   }
 
   log.blank();
+
+  if (data.searchFilterApplied) {
+    log.warn(
+      "Search results are filtered. To see all available libraries, update your public library access settings in the dashboard at https://context7.com/dashboard?tab=libraries"
```
fahreddinozcan: We don't mention how they're filtered, also these are already "all the available libraries to that user". maybe we should say sth like search over all public libraries? maybe

> Your results only include libraries matching your access settings. To search across all public libraries, update your settings at https://context7.com/dashboard?tab=libraries

fahreddinozcan (approved):

PR #2192, opened 2026-03-11
feat(cli): enumerate popularity and trust in skill search results

fahreddinozcan:
## Summary
- Replace raw install counts with star-based popularity rating (★☆☆ / ★★☆ / ★★★) based on install count thresholds (<100, <1000, 1000+)
- Replace numeric trust scores (0-10) with High/Medium/Low labels using MCP reputation thresholds (>=7, >=4, <4)
- Show install count range and raw trust score in the hover/description section
- Reorder `ctx7 setup` to show CLI + Skills before MCP server


packages/cli/src/utils/prompts.ts
```diff
@@ -14,19 +14,56 @@ export function terminalLink(text: string, url: string, color?: (s: string) => s
...
+export function formatPopularity(count: number | undefined): string {
+  const filled = "★";
+  const empty = "☆";
+  const max = 4;
+  let stars: number;
+  if (count === undefined || count === 0) stars = 0;
+  else if (count < 100) stars = 1;
+  else if (count < 500) stars = 2;
+  else if (count < 1000) stars = 3;
+  else stars = 4;
```
enesgules, line 29: there are max 4 stars but when a library has less, it shows 1 start and 2 empty stars so 3 in total

enesgules (approved):

packages/cli/src/utils/prompts.ts
fahreddinozcan: if it's larger than 1000, it shows 4 stars

image

PR #2271, opened 2026-03-18
fix(mcp): support NODE_EXTRA_CA_CERTS for enterprise MITM proxies

### LINKED ISSUE #2268: [Bug]: `TypeError: fetch failed` behind enterprise MITM proxies (NODE_EXTRA_CA_CERTS ignored)
### Bug Description

When running `@upstash/context7-mcp` behind an enterprise transparent MITM proxy (like Zscaler or corporate VPNs), the server throws `TypeError: fetch failed` due to an `UNABLE_TO_VERIFY_LEAF_SIGNATURE` TLS error. 

While users attempt to fix this by setting the standard `NODE_EXTRA_CA_CERTS` environment variable, the setting is frequently stripped, ignored, or injected too late in the lifecycle when the MCP server is spawned by background client daemons (like Gemini CLI or certain Cursor configs) via `npx`.

Because the current `api.ts` relies strictly on Node's default `fetch()` TLS context (which is frozen at binary startup), the custom corporate certificates are ignored, severing the connection to `mcp.context7.com`.

### Proposed Fix

The fix I think should be simple. You are already utilizing `setGlobalDispatcher` from `undici` to handle explicit HTTP proxies. You can use that exact same mechanism to explicitly inject the user's CA certificates into the global dispatcher at runtime, completely bypassing the environment-stripping issues of MCP clients.

In `packages/mcp/src/lib/api.ts`, right around the proxy injection logic, you can add:

```typescript
import { Agent, ProxyAgent, setGlobalDispatcher } from "undici";
import * as fs from "fs";

// ... existing PROXY_URL logic ...

// If no explicit proxy URL is set, but custom CA certs are provided (for transparent proxies)
if (!PROXY_URL && process.env.NODE_EXTRA_CA_CERTS) {
  try {
    const ca = fs.readFileSync(process.env.NODE_EXTRA_CA_CERTS);
    setGlobalDispatcher(new Agent({ connect: { ca } }));
  } catch (error) {
    console.error(`[Context7] Failed to load custom CA certificates:`, error);
  }
}
```
*(If `PROXY_URL` is set, the `ca` option should also be merged into the `ProxyAgent` options).*

### Steps to Reproduce
1. Connect to an enterprise network with a transparent SSL intercept proxy.
2. Configure an MCP client to launch `npx -y @upstash/context7-mcp`.
3. Provide `NODE_EXTRA_CA_CERTS=/path/to/corporate/cert.pem` in the MCP environment config.
4. Attempt to resolve a library.
5. Observe `TypeError: fetch failed`.

### Environment
- Node.js v18+ (Specifically reproduced on v24.14.0)
- `@upstash/context7-mcp` via `npx`
- Enterprise Transparent Proxy (MITRE / Zscaler)

mvanhorn:
## Summary

- Reads custom CA certificates from `NODE_EXTRA_CA_CERTS` and injects them into undici's global dispatcher at runtime
- Handles both transparent proxy (no explicit HTTPS_PROXY) and explicit proxy scenarios
- Leverages the existing `setGlobalDispatcher` + `undici` pattern already used for HTTP proxy support

## Problem

Behind enterprise MITM proxies (Zscaler, corporate VPNs), Node's default `fetch()` TLS context does not trust the corporate CA certificate. Even when `NODE_EXTRA_CA_CERTS` is set, many MCP client launchers (npx, background daemons) strip or inject the env var too late for Node to pick it up at startup, causing `UNABLE_TO_VERIFY_LEAF_SIGNATURE` errors.

## Solution

Explicitly read the CA cert file at module load time and pass it to undici's `Agent` or `ProxyAgent` via the `connect.ca` option. This bypasses Node's frozen TLS context and works regardless of when the env var was set.

Three cases are handled:
1. **Explicit proxy + custom CA:** `ProxyAgent` gets both the proxy URI and the CA cert
2. **No proxy + custom CA (transparent MITM):** A plain `Agent` with the CA cert is set as the global dispatcher
3. **No custom CA:** Existing behavior is unchanged

Fixes #2268

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/.changeset/brave-foxes-dance.md
+++ b/.changeset/brave-foxes-dance.md
@@ -1,0 +1,5 @@
+---
+"@upstash/context7-mcp": patch
+---
+
+Support NODE_EXTRA_CA_CERTS for enterprise MITM proxies by injecting custom CA certificates into undici's global dispatcher at runtime
--- a/packages/mcp/src/lib/api.ts
+++ b/packages/mcp/src/lib/api.ts
@@ -1,7 +1,8 @@
 import { SearchResponse, ContextRequest, ContextResponse } from "./types.js";
 import { ClientContext, generateHeaders } from "./encryption.js";
-import { ProxyAgent, setGlobalDispatcher } from "undici";
+import { Agent, ProxyAgent, setGlobalDispatcher } from "undici";
 import { CONTEXT7_API_BASE_URL } from "./constants.js";
+import { readFileSync } from "fs";
 
 /**
  * Parses error response from the Context7 API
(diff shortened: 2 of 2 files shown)


fahreddinozcan (approved):

mvanhorn:
Thanks for the quick review and merge on both PRs.

PR #2291, opened 2026-03-20
feat(mcp): add --version / -v flag to MCP CLI binary

### LINKED ISSUE #2284: [Feature]: add --version/-v to mcp cli
### Problem Description

Currently the help menu only displays the following content:
```shell
$ context7-mcp -h
Usage: index [options]

Options:
  --transport <stdio|http>  transport type (default: "stdio")
  --port <number>           port for HTTP transport (default: "3000")
  --api-key <key>           API key for authentication (or set CONTEXT7_API_KEY env var)
  -h, --help                display help for command
```

There is no `--version` or `-v` flag
The only way to check the version of the binary is to run it with no arguments and inspect the output string:
```shell
$ context7-mcp
Context7 Documentation MCP Server v2.1.4 running on stdio
```

### Proposed Solution

Add a `--version` and/or `-v` flag and print out the raw version value

### Alternatives Considered

_No response_

### Priority

Would improve my workflow

### Additional Context

I mainly request this, so I can package it nicely into nixpkgs with a test version hook. This will make it easier to test if the binary still runs if some changes happen in the future and catch it before it goes into production!

withsivram:
## Summary

- Wires commander's `.version()` to `SERVER_VERSION` (already read from `package.json`) so that running `npx @upstash/context7-mcp --version` or `-v` prints the current package version and exits cleanly
- One-line change in `packages/mcp/src/index.ts`

## Test plan

- [ ] `npx @upstash/context7-mcp --version` prints version number and exits
- [ ] `npx @upstash/context7-mcp -v` same behavior
- [ ] Normal MCP server startup is unaffected

Closes #2284

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -18,6 +18,7 @@ const DEFAULT_PORT = 3000;
 
 // Parse CLI arguments using commander
 const program = new Command()
+  .version(SERVER_VERSION, "-v, --version", "output the current version")
   .option("--transport <stdio|http>", "transport type", "stdio")
   .option("--port <number>", "port for HTTP transport", DEFAULT_PORT.toString())
   .option("--api-key <key>", "API key for authentication (or set CONTEXT7_API_KEY env var)")


enesgules:
This is already in master so closing. thanks for the pr

PR #2320, opened 2026-03-25
feat(setup): install rules alongside skills, update rule/skill content for better trigger rates

fahreddinozcan:
## Summary

Based on invocation routing eval benchmarks (60 queries, 6 modes, clean + mid-session context):

- **Skill-only**: 66% recall
- **Skill + rule**: 96-98% recall
- **MCP + rule**: 98-100% recall

## Changes

1. **Rule templates updated** to be selective with explicit should/should-not examples and library names. Split into `MCP_RULE_CONTENT` and `CLI_RULE_CONTENT` variants.

2. **CLI setup now installs a rule** alongside the skill. Previously `ctx7 setup --cli` only installed the find-docs skill (66% trigger rate). Now it also writes a rule file for each selected agent.

3. **find-docs skill description updated** from passive ("Prefer this skill whenever documentation accuracy matters") to pushy ("Use even when you think you know the answer -- verify against current docs"). Jumps recall from 66% to 98%.

4. **Removed `alwaysApply` frontmatter** from rule templates -- not a Claude Code feature.

## Eval report

See `skills/find-docs-workspace/REPORT.md` on the eval branch for full methodology and results.

### DIFF
--- a/rules/context7-cli.md
+++ b/rules/context7-cli.md
@@ -1,6 +1,6 @@
 Use the `ctx7` CLI to fetch current documentation before answering questions about libraries, frameworks, or APIs -- even well-known ones like React, Next.js, Prisma, Express, Tailwind, Django, or Spring Boot. Your training data may not reflect recent changes.
 
-IMPORTANT: Use even when you think you know the answer. Verify against current docs.
+IMPORTANT: Use even when you think you know the answer. Verify against current docs. If you're about to search the web for library documentation or API details, use Context7 first.
 
 Always use for: API syntax questions, configuration options, version migration issues, "how do I" questions mentioning a library name, debugging that involves library-specific behavior, setup instructions, and CLI tool usage.
 
--- a/rules/context7-cli.md
+++ b/rules/context7-cli.md
@@ -9,6 +9,8 @@ Do not use for: refactoring, writing scripts from scratch, debugging business lo
 3. Fetch docs: `npx ctx7@latest docs <libraryId> "<user's question>"`
 4. Answer using the fetched documentation
 
-Use the user's full question as the query, not single words. Do not run more than 3 commands per question.
+You MUST call `library` first to get a valid ID (format: `/org/project`) unless the user provides one directly. Use the user's full question as the query -- specific and detailed queries return better results than vague single words. Do not run more than 3 commands per question. Do not include sensitive information (API keys, passwords, credentials) in queries.
 
-For details on authentication, version-specific IDs, and library resolution, see the `find-docs` skill if available.
+For version-specific docs, use `/org/project/version` from the `library` output (e.g., `/vercel/next.js/v14.3.0`).
+
+If a command fails with a quota error, inform the user and suggest `npx ctx7@latest login` or setting `CONTEXT7_API_KEY` env var for higher limits. Do not silently fall back to training data.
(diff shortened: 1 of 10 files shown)


enesgules (approved):

rules/context7-cli.md
```diff
@@ -0,0 +1,16 @@
+Use the `ctx7` CLI to fetch current documentation whenever the user asks about a library, framework, SDK, API, CLI tool, or cloud service -- even well-known ones like React, Next.js, Prisma, Express, Tailwind, Django, or Spring Boot. This includes API syntax, configuration, version migration, library-specific debugging, setup instructions, and CLI tool usage. Use even when you think you know the answer -- your training data may not reflect recent changes. Prefer this over web search for library docs.
+
+Do not use for: refactoring, writing scripts from scratch, debugging business logic, code review, or general programming concepts.
+
+## Steps
+
+1. Resolve library: `npx ctx7@latest library <name> "<user's question>"`
+2. Pick the best match by: exact name match, description relevance, code snippet count, source reputation (High/Medium preferred), and benchmark score (higher is better). If results don't look right, try the full name with punctuation (e.g., "next.js" not "nextjs")
+3. Fetch docs: `npx ctx7@latest docs <libraryId> "<user's question>"`
```
enesgules, line 9: Should we explain the format of libraryId here?

rules/context7-cli.md
```diff
@@ -0,0 +1,16 @@
...
+4. Answer using the fetched documentation
```
enesgules, line 10: Maybe we should suggest trying different queries?

rules/context7-mcp.md
```diff
@@ -0,0 +1,10 @@
+Use Context7 MCP to fetch current documentation whenever the user asks about a library, framework, SDK, API, CLI tool, or cloud service -- even well-known ones like React, Next.js, Prisma, Express, Tailwind, Django, or Spring Boot. This includes API syntax, configuration, version migration, library-specific debugging, setup instructions, and CLI tool usage. Use even when you think you know the answer -- your training data may not reflect recent changes. Prefer this over web search for library docs.
+
+Do not use for: refactoring, writing scripts from scratch, debugging business logic, code review, or general programming concepts.
+
+## Steps
+
+1. `resolve-library-id` with the library name and the user's question
```
enesgules: Should we say always start with resolve library id tool call if the user doesn't provide an exact libraryId in /org/repo format?

rules/context7-mcp.md
enesgules, line 1: This file has some duplicate content with our MCP server prompts but since it's a small file the amount it takes up in the context window should be negligable. But just wanted to point that out if it wasn't intended. Maybe instead of functionality and explaining how to use, we can point out the benefits of the MCP more?

skills/find-docs/SKILL.md
```diff
@@ -1,25 +1,18 @@
...
+  developer technology. Use this skill whenever the user asks about a specific
+  library, framework, SDK, CLI tool, or cloud service -- even for well-known ones
+  like React, Next.js, Prisma, Express, Tailwind, Django, or Spring Boot. Your
+  training data may not reflect recent API changes or version updates.
+
+  Always use for: API syntax questions, configuration options, version migration
+  issues, "how do I" questions mentioning a library name, debugging that involves
+  library-specific behavior, setup instructions, and CLI tool usage.
+
+  Use even when you think you know the answer -- verify against current docs.
```
enesgules: Maybe we can mention training data shouldn't be trusted more explicitly

PR #2441, opened 2026-04-11
fix: skill installation path check compatibility for Windows

Miraclexc:
The path security check in `installSkillFiles` was hardcoded to use forward slashes (`/`). 
On Windows systems, `path.resolve` returns backslashes (`\`), causing valid paths to be 
rejected with an "outside the target directory" error.

This PR adds a check for backslashes to ensure cross-platform compatibility.

### DIFF
--- a/.changeset/bright-pants-hug.md
+++ b/.changeset/bright-pants-hug.md
@@ -1,0 +1,5 @@
+---
+"ctx7": patch
+---
+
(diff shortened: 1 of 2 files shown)


fahreddinozcan (approved):

fahreddinozcan:
Thanks for the contribution!

PR #2629, opened 2026-05-15
feat(mcp): multi-tenant Entra ID validation

fahreddinozcan:
## Summary
- Detect inbound Entra v2 tokens by issuer pattern, fetch per-teamspace configuration from the Context7 app, and verify against the matching tenant's JWKS.
- MCP server only validates — user resolution happens in the app middleware against `entra_user_mappings`.
- Per-tenant JWKS cache + 5-minute in-memory config cache keyed by JWT audience.
- Docs split out to #2636 so they can land first.

## Test plan
- [ ] `pnpm --filter @upstash/context7-mcp test` passes (jwt.test.ts covers Entra path + Clerk path + scope enforcement)
- [ ] Hit MCP with an Entra v2 token through APIM — token validates and forwards to the app
- [ ] Hit MCP with an unknown-audience token — rejected with `Unknown audience`
- [ ] Hit MCP with a Clerk token — still works (existing path untouched)

### DIFF
--- a/packages/mcp/src/lib/jwt.ts
+++ b/packages/mcp/src/lib/jwt.ts
@@ -32,17 +32,26 @@ async function fetchEntraConfig(audience: string): Promise<EntraConfig | null> {
   const cached = configByAudience.get(audience);
   if (cached && cached.expiresAt > now) return cached.value;
 
-  let value: EntraConfig | null = null;
   try {
     const res = await fetch(
       `${CONTEXT7_API_BASE_URL}/v2/entra/config/${encodeURIComponent(audience)}`
     );
-    if (res.ok) value = (await res.json()) as EntraConfig;
+    if (res.ok) {
+      const value = (await res.json()) as EntraConfig;
+      configByAudience.set(audience, { value, expiresAt: now + CONFIG_TTL_MS });
+      return value;
+    }
+    if (res.status === 404) {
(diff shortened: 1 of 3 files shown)


packages/mcp/src/lib/jwt.ts
```diff
@@ -1,27 +1,87 @@
...
+  return jwks;
+}
+
+interface EntraConfig {
+  teamspaceId: string;
+  tenantId: string;
+  requiredScope: string | null;
+}
+
+const CONFIG_TTL_MS = 5 * 60 * 1000;
```
enesgules, line 27: Caching null results under the same 5-minute TTL means one transient failure (5xx, network blip, JSON parse error) locks every Entra token for that audience out of the server for 5 minutes with a misleading Unknown audience error.
Suggest only negative-caching on an explicit 404 - anything else should bypass the cache so the next request retries:
if (res.ok) {
    value = (await res-json()) as EntraConfig;
} else if (res status != 404) {
    return null; // transient - don't cache
}

In the app we have
 - 404 when getEntraConfigByAudience(audience) returns null — that's the authoritative "this audience is not configured" answer. Safe to cache.
  - 400 for missing audience (won't happen from the MCP server).
  - 500 in the catch block — any exception thrown by Redis or the service layer. That's the "I don't know" case. Not safe to cache.

enesgules (approved):

PR #2655, opened 2026-05-20
feat: add sessions to MCP server

enesgules:


packages/mcp/src/index.ts
```diff
@@ -387,22 +399,17 @@ async function main() {
...
-      // Reject GET requests — this server is stateless and does not send server-initiated
-      // notifications, so SSE streams serve no purpose and cause mass NGINX timeouts.
-      // Returning 405 is spec-compliant per MCP StreamableHTTP (2025-03-26).
-      if (req.method === "GET") {
-        return res.status(405).json({
-          jsonrpc: "2.0",
-          error: { code: -32000, message: "Server does not support GET requests" },
-          id: null,
-        });
-      }
```
fahreddinozcan, line 401: Why did you remove this? We should be returning 405 on this case, but I can't see such response in new code.

fahreddinozcan:
We should have the backend changes in place as well to not miss the session id logs

packages/mcp/src/index.ts
```diff
@@ -577,6 +604,19 @@ async function main() {
       });
     };
 
+    process.on("SIGINT", async () => {
```
fahreddinozcan: we need to handle the SIGTERM event too

packages/mcp/src/index.ts
```diff
@@ -387,22 +399,17 @@ async function main() {
       );
     };
 
+    // Session-keyed transport map. One StreamableHTTPServerTransport (and one connected
+    // McpServer) per MCP session, looked up on every request via the `mcp-session-id`
+    // header. Both `/mcp` and `/mcp/oauth` share this map — auth is enforced per request
+    // via `requireAuth`, not stored on the session.
+    const transports: { [sessionId: string]: StreamableHTTPServerTransport } = {};
```
fahreddinozcan: Currently we have a memory leak. This holds a plain object map and only deletes on transport.onclose. In practice MCP clients frequently drop without sending DELETE, laptop closes lid, network blip, process killed, etc. Those sessions sit in transports[] forever, and InMemoryEventStore.events grows unbounded with every tool result on every live session. We could keep track of a thing like `lastActivity` field and remove unused clients. If there is a native way to do this in sdk, that goes as well.

packages/mcp/src/index.ts
```diff
@@ -448,25 +455,45 @@ async function main() {
...
-        // because some MCP HTTP clients cap the underlying fetch at 60s waiting for
-        // headers, even though the per-tool timeout is much higher.
-        const transport = new StreamableHTTPServerTransport({
-          sessionIdGenerator: undefined,
-          enableJsonResponse: false,
-        });
+        const sessionId = req.headers["mcp-session-id"] as string | undefined;
+        let transport: StreamableHTTPServerTransport;
+
+        if (sessionId && transports[sessionId]) {
```
fahreddinozcan: This is wrong and would break things. Since sessions are kept in memory, if a request lands on a new pod(where this transport is not registered), they'd basically get 400 response. 

We need to:
* keep transports in redis
* make sure there's sticky sessions on on the ingress level.

packages/mcp/src/index.ts
```diff
@@ -10,8 +10,11 @@ import { formatSearchResults, extractClientInfoFromUserAgent } from "./lib/utils
 import { isJWT, validateJWT } from "./lib/jwt.js";
 import express from "express";
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
+import { isInitializeRequest } from "@modelcontextprotocol/sdk/types.js";
+import { InMemoryEventStore } from "@modelcontextprotocol/sdk/examples/shared/inMemoryEventStore.js";
```
fahreddinozcan: the comments of this basically say that persistent store would be better instead of this "example" event store. i think persistence would be nice.

fahreddinozcan:
Can we make sure that session id's are usable as trace ids? Because single session id doesn't really mean a single chat history, imo. A single long-lived MCP client (e.g. Claude Code session lasting hours) emits dozens of unrelated resolve-library-id / query-docs calls — they all share one session id. The trace will be very noisy.

.changeset/strict-feet-end.md
```diff
@@ -0,0 +1,8 @@
+---
+"@upstash/context7-mcp": minor
+---
+
+Add stateful sessions to the HTTP transport.
+
+The `--transport http` server now follows the MCP Streamable HTTP session model
+instead of constructing a fresh transport per request:
```
fahreddinozcan: is this truncated

fahreddinozcan:
Can we introduce some tests maybe?

fahreddinozcan:
2. Inconsistent fail-open vs fail-closed on Redis errors.
In [sessionStore.ts:18-22](vscode-webview://10j61ms2dm3pbsqcnmgeha12megubu06uqrmvmnbovnccv63nr13/packages/mcp/src/lib/sessionStore.ts#L18-L22), refresh() swallows the Redis error and returns true (fail-open). But create() (line 13) and delete() (line 26) let the error propagate, which in [index.ts:520-528](vscode-webview://10j61ms2dm3pbsqcnmgeha12megubu06uqrmvmnbovnccv63nr13/packages/mcp/src/index.ts#L520-L528) becomes a 500. Net effect: if Redis blips, an in-flight session keeps working, but a brand-new client cannot initialize and an explicit logout 500s. Either commit to fail-open everywhere or fail-closed everywhere — current shape gives the worst of both.

3. refresh() returning true on error silently disables session validation.
Once redis.expire throws, every non-initialize POST is treated as if the session is valid, regardless of whether the ID was ever issued. That makes session validation effectively decorative under Redis outages. If you want fail-open for availability, fine — but it should be a deliberate, commented decision, not buried in a generic catch.

fahreddinozcan (approved):

PR #2666, opened 2026-05-22
feat: pi context7 extension

enesgules:


packages/pi/skills/context7-docs/SKILL.md
fahreddinozcan, line 1: I think it might be possible to get the skill from the root `skills` folder, so that we only have single file to maintain. I guess it would be possible with build configs

packages/pi/prompts/c7-docs.md
```diff
@@ -0,0 +1,12 @@
+---
+description: Fetch Context7 documentation for a library
+argumentHint: <library> <question>
```
fahreddinozcan: as per spec, it should be like this
https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/prompt-templates.md#argument-hints
```suggestion
argument-hint: <library> <question>
```

packages/pi/lib/api.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (!response.ok) {
+    return parseErrorResponse(response);
+  }
+
+  const text = await response.text();
+  if (!text) {
+    return "Documentation not found or not finalized for this library. This might have happened because you used an invalid Context7-compatible library ID. To get a valid Context7-compatible library ID, use the 'resolve-library-id' with the package name you wish to retrieve documentation for.";
+  }
+  return text;
+}
```
fahreddinozcan, line 65: do we need try/catch for these? or does pi handle it

packages/pi/skills/context7-docs/SKILL.md
enesgules: We don't build anything, pi consumes the ts files. There are also subtle differences and we may go forward separately for now imo

packages/pi/lib/api.ts
enesgules: "Tool execute errors must be signaled by throwing; the thrown error is caught, reported to the LLM with isError: true, and execution continues." per https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/extensions.md#error-handling

fahreddinozcan (approved):

Issue #2811, opened 2026-06-22
[Bug]: Accidental case-sensitivity for case-insensitive identifiers


vitorsr:
### MCP Client

Other (specify in description)

### Context7 MCP Version

-

### Bug Description

GitHub usernames and repositories are case-insensitive but Context7 does not respect case insensitivity across various surfaces. Other documentation sources such as websites also have case-insensitive parts such as the domain name and TLD which are not respected.

Example (website front-end):

https://context7.com/ClickHouse/ClickHouse (note: ClickHouse with this capitalization is the canonical organization and repository names as returned by the GitHub API, yet it 404s).

Example (MCP):

```
{
  "libraryId": "/ClickHouse/ClickHouse",
  "query": "How to create a table in ClickHouse?"
}
```

```
Library "/ClickHouse/ClickHouse" not found. Please check the library ID or your access permissions.
```

Having to *always* resolve the identifier instead of querying against canonical names is wasteful and counterintuitive.

### Steps to Reproduce

See Bug Description.

### Expected Behavior

See Bug Description.

### Actual Behavior

See Bug Description.

### Error Messages / Logs

```shell

```

### Transport Method

http

### Configuration

```json

```

enesgules:
Hi, thanks for reporting, this is now fixed

PR #2812, opened 2026-06-23
fix: normalize context library IDs in clients

chaochaoweb3:
## Summary
- normalize the case-insensitive library ID segments before CLI docs lookups
- normalize the same segments before MCP context requests
- keep version/tag suffix casing unchanged

Addresses the CLI/MCP client side of #2811.

## Test
- pnpm --filter ctx7 test -- library-id.test.ts
- pnpm --filter ctx7 typecheck
- pnpm --filter @upstash/context7-mcp typecheck
- pnpm --filter ctx7 format:check
- pnpm --filter @upstash/context7-mcp format:check
- git diff --check


enesgules:
Hi, this no longer is needed since it's fixed on the API

PR #2894, opened 2026-07-10
docs: add Autohand Code MCP setup

igorcosta:
## What changed

Added Autohand Code to the existing MCP client installation list with both local npm and remote HTTP setup commands.

## Why

Context7 already documents a broad set of MCP clients. These commands give Autohand Code users the same direct path without changing the server or its existing integrations.

## Validation

- Verified the commands against the current Autohand Code MCP CLI implementation
- Ran `git diff --check`

### DIFF
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -89,6 +89,25 @@ You can find your Smithery key in the [Smithery.ai webpage](https://smithery.ai/
 </details>
 
 <details>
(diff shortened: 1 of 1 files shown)


xtwitter112233-maker (approved):

enesgules:
Hi, thanks! Could you move it down to the bottom of the file

enesgules (approved):

PR #2922, opened 2026-07-17
fix: extract x-context7-api-key header and tighten JWT detection

fix2015:
## Summary

- **Add missing `x-context7-api-key` header extraction**: The CORS config already allows the `X-Context7-API-Key` header, but `extractApiKey` never checked for it. Added `x-context7-api-key` to the extraction chain so clients sending that header are properly authenticated.

- **Fix `isJWT` false positive for strings with two dots**: The previous check `token.split(".").length === 3` returned `true` for strings like `".."` or `"a..b"`. Now it also verifies that all three parts are non-empty, preventing false positives from malformed tokens.

## Test plan

- Verify that requests with `X-Context7-API-Key` header are correctly authenticated
- Verify that `isJWT("..")` and `isJWT("a..b")` return `false`
- Verify that `isJWT("header.payload.signature")` still returns `true`

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -372,6 +372,7 @@ async function main() {
       return (
         extractBearerToken(req.headers.authorization) ||
         extractHeaderValue(req.headers["context7-api-key"]) ||
+        extractHeaderValue(req.headers["x-context7-api-key"]) ||
         extractHeaderValue(req.headers["x-api-key"]) ||
         extractHeaderValue(req.headers["context7_api_key"]) ||
         extractHeaderValue(req.headers["x_api_key"])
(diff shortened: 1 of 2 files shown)


enesgules:
is there a specific reason why you added it? a specific client in mind?

fix2015:
Hey @enesgules, good question! The main motivation is supporting clients that pass their API key via a custom header (x-context7-api-key) rather than the standard Authorization bearer token — for example, lightweight CLI tools or scripts where setting a custom header is simpler than constructing a full Bearer token flow.

The JWT detection tightening is a separate concern: the existing check only looked at string length, so a long non-JWT string could accidentally trigger JWT parsing. The updated check validates the three-segment dot-separated structure that all JWTs have.

enesgules:
hey, closing this since it doesn't provide too much value. thanks for the pr

PR #2942, opened 2026-07-24
docs: add mcpindex screened badge

gautamgb:
Hey, Context7 is one of the MCP servers we see referenced constantly, so figured this was worth doing properly rather than assuming. mcpindex ran a semantic screen against your registered server and it came back clean (no injection/manipulation patterns found in the tool description); screened is advisory, not a security cert, just an honest description-level check. Added the badge to your README in case it's useful. No worries at all if you'd rather close this.


enesgules:
Thanks but we prefer to keep our readme relatively small so closing. Thanks for creating the pr

PR #2958, opened 2026-07-29
CTX7-1764: lazy authentication on /mcp

fahreddinozcan:
Anonymous clients keep working; the server issues an OAuth challenge only when an unauthenticated caller hits a protected tool or spends its anonymous allowance. Needs upstash/context7app#821 (removes the sign-in prompt header this replaces).

- Gate `tools/call` at the HTTP layer, before the transport streams a 200 — `initialize` and `tools/list` always pass
- Triggers: `CONTEXT7_PROTECTED_TOOLS` (by name) and a per-client Redis quota, `CONTEXT7_ANON_FREE_CALLS`, default 5, fail-open
- Challenge shape per client family: HTTP 401 + `WWW-Authenticate` by default, `_meta["mcp/www_authenticate"]` on a failed `CallToolResult` for ChatGPT/Codex, which ignore a bare 401
- Advertise per-tool `securitySchemes` in `tools/list`; injected by wrapping the SDK handler, since `registerTool` drops unknown config keys
- Serve the RFC 9728 path-suffixed PRM variants and `resource_documentation`
- Drop the anonymous sign-in elicitation and the `X-Context7-Auth-Prompt` reader

## Testing

`scripts/lazy-auth-probe.mjs` drives both client families over raw HTTP and prints the status, header and `_meta` each one sees:

```
docker compose -f scripts/docker-compose.redis.yml up -d
UPSTASH_REDIS_REST_URL=http://localhost:8079 UPSTASH_REDIS_REST_TOKEN=local_token \
  CONTEXT7_ANON_FREE_CALLS=3 node dist/index.js --transport http --port 3000
node scripts/lazy-auth-probe.mjs
```

Ran against a local server: spec client gets `401` + `Bearer error="invalid_token", error_description=…, resource_metadata=…, scope="profile email"` on call 4; Codex UA gets `200` carrying the same challenge in `_meta`; both PRM paths resolve. 71 unit tests pass.


fahreddinozcan:
Moved to CTX7-1886: upstash/context7#2959

PR #2967, opened 2026-07-31
test: verify MiniMax models through Context7Agent

octo-patch:
Reason: Add a MiniMax OpenAI-compatible fixture that verifies both current models through Context7Agent at the configured regional endpoints.

- Add `@ai-sdk/openai` as a development dependency.
- Add a deterministic fixture test for `MiniMax-M3` and `MiniMax-M2.7` at the global and China endpoints.
- Verify that the Context7Agent path sends chat-completion requests with the selected model and Context7 tools.

Checks:
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm --filter @upstash/context7-tools-ai-sdk typecheck`
- `pnpm --filter @upstash/context7-tools-ai-sdk exec vitest run src/minimax.test.ts`
- `./node_modules/.bin/tsup` from `packages/tools-ai-sdk`
- ESLint and Prettier checks for the changed files
- Secret scan passed
- The full package suite was also run; its five network-backed cases require an unset cloud-region environment variable, while the remaining fifteen tests passed.

### DIFF
--- a/packages/tools-ai-sdk/package.json
+++ b/packages/tools-ai-sdk/package.json
@@ -38,6 +38,7 @@
     "ai": "^6.0.5",
     "zod": "^4.4.3",
     "@ai-sdk/amazon-bedrock": "^4.0.9",
+    "@ai-sdk/openai": "^3.0.88",
     "@types/node": "^25.0.3",
     "dotenv": "^17.4.2",
     "tsup": "^8.5.1",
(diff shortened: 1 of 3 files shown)


fahreddinozcan:
Hey @octo-patch, what's the motivation of this test? Is there a problem with minimax models on our side?

octo-patch:
There isn't a known MiniMax-specific problem in Context7. This is deterministic compatibility coverage that verifies Context7Agent accepts the OpenAI-compatible MiniMax provider for both model IDs and regional base URLs, and forwards Context7 tools through the normal agent path without making live API calls.

PR #3008, opened 2026-08-10
feat(plugins): add Context7 plugin for OpenCode

fahreddinozcan:
Adds `@upstash/context7-opencode`, an OpenCode plugin that wires up Context7 in one command.

- New package `packages/opencode`, published as `@upstash/context7-opencode`, installed with `opencode plugin @upstash/context7-opencode`
- Registers the Context7 MCP server and the `context7-mcp` skill through the plugin `config` hook
- `/mcp/oauth` when no key is present, `/mcp` with an `Authorization` header when `CONTEXT7_API_KEY` or the `apiKey` option is set
- Both entries are additive and skipped when the user already configured the same key
- The bundled skill is a verbatim copy of `skills/context7-mcp/SKILL.md`, matching the claude, copilot, codex and cursor plugin copies
- `docs/clients/opencode.mdx` keeps `npx ctx7 setup --opencode` as the primary install path and lists the plugin as an alternative

Merging publishes a new package at `0.1.0` via the changeset.

## Why an npm package

OpenCode has no directory-based plugin marketplace like Claude Code, Copilot CLI, or Codex. `opencode plugin <module>` resolves npm specs, and git specs cannot point at a subdirectory of a monorepo, so a `github:upstash/context7` spec can never reach `packages/opencode`. This follows `packages/pi`, which is an npm package for the same reason.

The plugin is a server-target module with a default `{ id, server }` export, which is what OpenCode's loader reads. It deliberately exports nothing else: any additional export is picked up as a second plugin by the legacy loader. Server plugins do not appear in the TUI Plugins dialog, which lists TUI-target plugins only.

## Verified

There are no unit tests. The package is config wiring, so it was verified end to end against real OpenCode instead: published to a local registry, installed with `opencode plugin @upstash/context7-opencode` into a clean project, then run against a live model.

```
⚙ context7_resolve-library-id {"libraryName":"Next.js","query":"how to define middleware in Next.js 15"}
⚙ context7_query-docs {"libraryId":"/vercel/next.js/v15.1.8", ...}
```

It resolved the library, picked the version-specific ID on its own, and answered from real documentation. Checked with `claude-haiku-4-5` as well as `claude-sonnet-4-5`.

Also confirmed:

- Install detects the server target and writes `.opencode/opencode.json`
- `opencode debug skill` resolves `context7-mcp` from inside the installed package
- Both auth paths connect: OAuth on `/mcp/oauth` after `opencode mcp auth`, and an `Authorization` header on `/mcp`
- An existing `context7` MCP entry in the user config is left untouched
- Running both `ctx7 setup` and the plugin is safe, since OpenCode dedupes skills by name

## Note for reviewers

The branch history contains reversals from iterating on this: an agent, a slash command, unit tests, option toggles and a stdio transport were each added and then removed. The commit message on `7aa78fa` also records a "known limitation" about a 405 blocking the remote transport. That diagnosis was wrong, caused by a synthetic test environment with an empty HOME, and it does not reproduce on a normal installation. Squash-merging drops all of that.


enesgules (approved):
Tested locally: loaded `packages/opencode` as a path plugin spec in my global `opencode.json`.

- `opencode debug config` shows the hook injecting `mcp.context7` -> `/mcp/oauth` and the package `skills` path
- MCP dialog: `context7 needs_auth` -> connected after `opencode mcp auth`
- My pre-existing `context7` entry was left untouched, as documented; the plugin only registers when the key is free

LGTM.

Issue #3060, opened 2026-08-23
CLI setup silently drops pinned @upstash/context7-mcp version on re-setup when Codex config.toml arrays are multi-line or have trailing commas (readTomlServerEntry drops args)


wallidsaydi-creator:
## Summary

`readTomlServerEntry` in `packages/cli/src/setup/mcp-writer.ts` parses each `key = value` line of the `[mcp_servers.<name>]` block with `JSON.parse` **one line at a time**. Two classes of valid TOML therefore fail to parse and the key is **silently dropped** from the entry:

1. **Multi-line arrays** — `args = [` … `]` split across lines (what taplo, cargo-fmt, Prettier-TOML, and many editors produce once a line exceeds the print width — which any real api-key does)
2. **Trailing commas** — TOML permits `["a", "b",]`; JSON does not (breaks even single-line arrays)

## Impact

This breaks the documented stdio-preservation path in `resolveEntryToWrite` (`commands/setup.ts`): when an existing stdio `@upstash/context7-mcp` entry is detected, setup preserves the user's pinned version and flags and only swaps `--api-key`. But detection goes through `isStdioContext7Entry(existingEntry)`, which needs `args` to contain the package name. With `args` silently missing, detection fails and setup **falls back to a fresh canonical entry** — silently wiping a user's pinned version (e.g. `@upstash/context7-mcp@0.6.0` → floating `@upstash/context7-mcp`), custom `cwd`, extra flags, or env settings on re-setup.

The failure is silent on both ends: setup reports `reconfigured`, the user's pin is gone.

## Repro (verified on main @ c324828)

`~/.codex/config.toml` (taplo/Prettier-formatted, api-key makes the array exceed print width):

```toml
model = "gpt-5.2-codex"

[mcp_servers.context7]
command = "npx"
args = [
  "-y",
  "@upstash/context7-mcp@0.6.0",
  "--api-key",
  "ctx7sk_live_OLD_old_old_old_old_old_old",
]

[mcp_servers.filesystem]
command = "uvx"
args = ["mcp-server-filesystem", "/tmp"]
```

Run `ctx7 setup --codex --stdio` (e.g. to rotate the API key). Expected: pin `@0.6.0` preserved, key swapped. Actual config afterwards:

```toml
[mcp_servers.context7]
command = "npx"
args = ["-y","@upstash/context7-mcp","--api-key","ctx7sk_live_NEW"]
```

[...]

PR #3063, opened 2026-08-23
Fix Codex TOML parsing dropping multi-line/trailing-comma args arrays (silent version wipe on re-setup)

wallidsaydi-creator:
Follow-up to #3060 (thanks @fahreddinozcan for the quick ack). Opening this PR so the fix is reviewable directly — no pressure on timing.

## What it fixes

`readTomlServerEntry` in `packages/cli/src/setup/mcp-writer.ts` parses Codex `config.toml` server entries with per-line `JSON.parse`, which silently drops `args` when:

1. **The array is multi-line** — what taplo/Prettier/cargo-fmt produce once a line exceeds print width (any real API key makes this likely).
2. **The array has a trailing comma** — valid TOML, invalid JSON.

Downstream, `isStdioContext7Entry` then fails, `resolveEntryToWrite` falls back to a fresh canonical entry, and a re-setup that was only meant to rotate the API key **silently wipes the user's pinned `@upstash/context7-mcp@<version>` to a floating version**. Full repro chain in #3060.

## The change

- String-aware bracket balancing to join multi-line arrays before parsing.
- String-aware trailing-comma strip (commas inside string literals are untouched).

## Verification

- Repro cases from #3060: 3/3 pass with the patch, fail on unpatched main (git-stash control).
- Edge cases: multi-line array with embedded commas in strings; trailing-comma single-line array — both parse correctly.
- Full CLI suite: **254/254 passing**; `tsc --noEmit` clean.

### DIFF
--- a/packages/cli/src/setup/mcp-writer.ts
+++ b/packages/cli/src/setup/mcp-writer.ts
@@ -119,10 +119,69 @@ export async function readTomlServerExists(filePath: string, serverName: string)
 /**
  * Reads the top-level `[mcp_servers.<serverName>]` block from a TOML config
  * file and parses its key-value lines into a JS object. Handles string and
- * array values (TOML array syntax is JSON-compatible). Sub-tables like
+ * array values (TOML array syntax is JSON-compatible), including arrays
+ * formatted across multiple lines. Sub-tables like
  * `[mcp_servers.<serverName>.http_headers]` are ignored. Returns undefined
  * if the file or section is missing.
  */
+/**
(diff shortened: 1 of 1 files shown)


wallidsaydi-creator:
Closing — superseded by #3075, which fixes the same defect more thoroughly (dedicated TOML string-array editor with fail-closed semantics rather than my minimal bracket-joining patch). Verified the merged fix covers both repro cases from #3060 (multi-line arrays and trailing commas). Thanks for the fast turnaround on the original report!

PR #3068, opened 2026-08-24
fix(mcp): validate opaque credentials before OAuth sessions

fahreddinozcan:
## Summary

- reject bearer values that do not match a supported Context7 API-key or OAuth-token format
- validate recognized opaque credentials through Context7's existing authenticated whoami endpoint before entering the MCP handler
- fail closed before MCP initialization, tool listing, or tool calls when validation fails or is unavailable
- use short bounded positive/negative caches keyed by credential hashes
- preserve JWT validation, valid OAuth/API-key access, and anonymous access through /mcp

## Security

Fixes the authentication bypass reported in Cobalt finding PT38659_11. Previously any bearer value that was not shaped like a JWT skipped validation and reached the MCP handler. Format recognition is only an early rejection gate: supported opaque formats still require explicit backend approval, so a well-shaped fake API key is rejected with 401 before MCP initialization.

Linear: CTX7-2021 (child of CTX7-2019)

## Validation

- pnpm run typecheck
- pnpm run lint:check
- pnpm test (6 files, 102 tests)
- Prettier check on changed TypeScript files


fahreddinozcan:
Closing as superseded. The downstream API already remains the authorization boundary for opaque credentials, and this PR duplicates opaque-token validation without completing MCP audience/resource validation or eliminating bearer-token passthrough. Replacement work will migrate Clerk MCP OAuth to JWT access tokens with strict issuer, audience, expiry, token-type, and scope validation, plus a documented rollout.

PR #3085, opened 2026-08-26
fix(cli): recover Scoop Git Bash library IDs

### LINKED ISSUE #3082: [Bug]: CLI fails to recover Git Bash-mangled library ID when Git is installed via Scoop
### MCP Client

Claude Code

### Context7 MCP Version

`ctx7` CLI 0.5.8 — CLI only; the MCP server is unaffected.

### Bug Description

On Windows, when Git is installed via Scoop, running `ctx7 docs /owner/repo "<query>"` in Git Bash fails with `Library ... not found`.

Git Bash applies MSYS2's automatic Unix→Windows path conversion to arguments passed to native binaries ([MSYS2 docs](https://www.msys2.org/docs/filesystem-paths/)): `/vercel/next.js` reaches Node as `D:/Scoop/apps/git/2.54.0/vercel/next.js`. Scoop exposes apps through a `current` junction, but the MSYS runtime resolves it to the real version dir when deriving its install root, so the mangled prefix is always `<scoop>/apps/git/<version>/` and the version segment changes on every Git upgrade. The recovery heuristic `recoverLibraryId` matches the `git` directory, but Scoop's layout inserts a `<version>/` segment, so the regex captures it as part of the ID and "recovers" `/2.54.0/vercel/next.js`. That string passes the format check `/^\/[^/]+\/[^/]/`, so the `Invalid library ID` error with the `//`-prefix hint never triggers, and the API returns not-found for the wrong ID.

Follow-up to #2733: that fix covers the standard install dirs but misses Scoop's `git/<version>/` layout.

### Steps to Reproduce

1. Install Git via Scoop on Windows: `scoop install git` — installs to `<scoop root>\apps\git\<version>` with a `current` junction
2. Open Git Bash — the shell Claude Code uses on Windows
3. Run `npx -y ctx7@latest docs /vercel/next.js "app router"`

Reproducible with a clean `bash --norc --noprofile` and no MSYS-related environment variables set — a default Scoop install, not a local customization.

### Expected Behavior

Docs for `/vercel/next.js` are returned.

### Actual Behavior

```
✖ Library "/2.54.0/vercel" not found. Please check the library ID or your access permissions.
```

### Error Messages / Logs

```shell
# End-to-end failure:
$ npx -y ctx7@latest docs /vercel/next.js "app router"
✖ Library "/2.54.0/vercel" not found. Please check the library ID or your access permissions.

# Both workarounds succeed:
$ npx -y ctx7@latest docs //vercel/next.js "app router"                        # OK
$ MSYS2_ARG_CONV_EXCL='*' npx -y ctx7@latest docs /vercel/next.js "app router" # OK
```

Relevant code, `packages/cli/src/utils/library-id.ts` v0.5.8:

```ts
const match = normalized.match(/^[A-Za-z]:\/.*?\/(?:Git|PortableGit|git-bash)\/(.+)$/i);
return match ? `/${match[1]}` : input;
```

Against `D:/Scoop/apps/git/2.54.0/vercel/next.js` this captures `2.54.0/vercel/next.js`.

### Transport Method

stdio (default) — N/A for the CLI.

### Node.js Version

v24.19.0

### Operating System

Windows 11 Pro (build 26200); Git 2.54.0.windows.1 via Scoop

### Configuration

N/A — CLI invoked directly: `npx -y ctx7@latest docs /vercel/next.js "app router"`

### Additional Context

- Workarounds, both verified:
  1. Extra leading slash: `ctx7 docs //vercel/next.js "<query>"` — the CLI's own hinted workaround: the invalid-ID error message in `packages/cli/src/commands/docs.ts` suggests it and `recoverLibraryId` collapses `//`, but that error path never triggers in this case.
  2. `MSYS2_ARG_CONV_EXCL='*' ctx7 docs /vercel/next.js "<query>"`.
- The MCP server is unaffected: `@upstash/context7-mcp` receives `libraryId` as a JSON tool parameter over MCP JSON-RPC, see the `query-docs` inputSchema in `packages/mcp/src/index.ts`; `process.argv` is only used for server flags such as `--port` and `--api-key`, so library IDs never cross the shell boundary where conversion happens.

fahreddinozcan:
## Summary

- recognize Scoop Git installs under `apps/git/<version>` when recovering Git Bash-mangled arguments
- remove Scoop's Git version segment without changing standard Git install behavior
- add regression coverage for slash styles, Git version suffixes, and version-like library owners

## Tests

- `pnpm --filter ctx7 test`
- `pnpm --filter ctx7 format:check`
- `pnpm --filter ctx7 build`

Closes #3082

### DIFF
--- a/packages/cli/src/utils/library-id.ts
+++ b/packages/cli/src/utils/library-id.ts
@@ -17,6 +17,14 @@ export function recoverLibraryId(input: string): string {
 
   // Strip the Git install dir, keeping the "/owner/repo[/version]" tail.
   const normalized = input.replace(/\\/g, "/");
+
+  // Scoop installs Git under apps/git/<version>. MSYS resolves Scoop's "current"
+  // junction, so discard that version segment before recovering the library ID.
+  const scoopMatch = normalized.match(
+    /^[A-Za-z]:\/.*\/apps\/git\/v?\d+(?:\.\d+)+(?:[^/]*)\/(.+)$/i
+  );
+  if (scoopMatch) return `/${scoopMatch[1]}`;
+
   const match = normalized.match(/^[A-Za-z]:\/.*?\/(?:Git|PortableGit|git-bash)\/(.+)$/i);
   return match ? `/${match[1]}` : input;
 }
--- a/packages/cli/src/utils/library-id.ts
+++ b/packages/cli/src/utils/library-id.ts
@@ -18,10 +18,10 @@ export function recoverLibraryId(input: string): string {
   // Strip the Git install dir, keeping the "/owner/repo[/version]" tail.
   const normalized = input.replace(/\\/g, "/");
 
-  // Scoop installs Git under apps/git/<version>. MSYS resolves Scoop's "current"
-  // junction, so discard that version segment before recovering the library ID.
+  // Scoop installs Git under apps/git/<version> with a "current" junction.
+  // Discard either form before recovering the library ID.
   const scoopMatch = normalized.match(
-    /^[A-Za-z]:\/.*\/apps\/git\/v?\d+(?:\.\d+)+(?:[^/]*)\/(.+)$/i
+    /^[A-Za-z]:\/.*\/apps\/git\/(?:v?\d+(?:\.\d+)+[^/]*|current)\/(.+)$/i
   );
   if (scoopMatch) return `/${scoopMatch[1]}`;
 
--- a/.changeset/cli-scoop-git-bash-path.md
+++ b/.changeset/cli-scoop-git-bash-path.md
@@ -1,0 +1,5 @@
+---
+"ctx7": patch
+---
+
+Recover library IDs mangled by Git Bash when Git is installed through Scoop.
--- a/packages/cli/src/__tests__/library-id.test.ts
+++ b/packages/cli/src/__tests__/library-id.test.ts
@@ -26,6 +26,17 @@ describe("recoverLibraryId", () => {
     expect(recoverLibraryId("D:/tools/PortableGit/facebook/react")).toBe("/facebook/react");
   });
 
+  test("recovers from a versioned Scoop Git install", () => {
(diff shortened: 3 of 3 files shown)


packages/cli/src/utils/library-id.ts
```diff
@@ -17,6 +17,14 @@ export function recoverLibraryId(input: string): string {
 
   // Strip the Git install dir, keeping the "/owner/repo[/version]" tail.
   const normalized = input.replace(/\\/g, "/");
+
+  // Scoop installs Git under apps/git/<version>. MSYS resolves Scoop's "current"
+  // junction, so discard that version segment before recovering the library ID.
+  const scoopMatch = normalized.match(
+    /^[A-Za-z]:\/.*\/apps\/git\/v?\d+(?:\.\d+)+(?:[^/]*)\/(.+)$/i
```
enesgules: If a path ever arrives with the `current` junction unresolved (`.../apps/git/current/vercel/next.js`), this regex does not match and the generic one returns `/current/vercel/next.js`. If MSYS always resolves the junction, ignore this; otherwise `(?:v?\d+(?:\.\d+)+[^/]*|current)` covers it.

enesgules: Nit: `(?:[^/]*)` — the group has no purpose, plain `[^/]*` is equal.

enesgules (approved):
LGTM — traced the edge cases (version-like owner, `apps/git` as owner/repo, version-pinned IDs through Scoop) and they all resolve correctly. Two minor notes inline, neither blocks merge.

packages/cli/src/utils/library-id.ts
fahreddinozcan: Handled in 147135b: the Scoop matcher now accepts both a version directory and an unresolved `current` junction. Added a focused regression test for `D:/Scoop/apps/git/current/vercel/next.js`.

fahreddinozcan: Handled in 147135b: removed the redundant non-capturing group and kept the suffix as plain `[^/]*`.

PR #1038, opened 2025-11-18
Add Emdash integration section to README

arnestrickmann:
This PR adds documentation for Emdash, an orchestration layer for running multiple coding agents in parallel.

**What this adds:**
- Documentation for Emdash's Context7 MCP support
- Explanation of global toggle and per-workspace enable features
- Reference to existing MCP client configuration sections
- Link to Emdash repository

Emdash allows users to run multiple coding agents (Codex, Claude Code, Cursor, etc.) in parallel with built-in Context7 MCP support. Users still need to configure their underlying agents to connect to Context7 MCP, which is documented in the existing sections above.

Issue #699, opened 2025-09-19
Refresh https://context7.com/workleap/wl-web-configs


patricklafrance:
I would appreciate if https://context7.com/workleap/wl-web-configs could be refreshed 🙏🏻 

Patrick

enesgules:
Started the process

PR #3088, opened 2026-08-27
fix(mcp): authenticate forwarded client IP assertions

fahreddinozcan:
## Summary

- let Express derive `req.ip` through a private/local trusted-proxy boundary
- delete the custom forwarded-header parser and duplicated IP-classification tests
- validate IP literals before signing short-lived AES-256-GCM assertions
- use a dedicated assertion key and retain the old key only for rollout compatibility
- decrypt the real emitted assertion in the spawned-server integration suite and verify the exact resolved address

Companion change for [CTX7-2529](https://linear.app/upstash/issue/CTX7-2529/prevent-unauthenticated-client-ip-override-from-bypassing-and).

Companion PRs: [API verifier #1038](https://github.com/upstash/context7app/pull/1038), [deployment #699](https://github.com/upstash/context7parser/pull/699).

## Verification

- MCP typecheck passed
- full MCP suite: 51 passed
- real local MCP + Next servers confirmed invalid IPs are not signed and valid/tampered/plaintext paths resolve safely

## Rollout

1. Provision both the new assertion secret and existing legacy secret in the deployment.
2. Deploy this producer and deployment PR first.
3. Deploy the API verifier.
4. Complete [CTX7-2536](https://linear.app/upstash/issue/CTX7-2536/remove-legacy-mcp-client-ip-encryption-after-assertion-rollout).


enesgules (approved):
Approve with suggestions. Root-cause fix: leftmost-XFF pick and plaintext fallback are gone, hardcoded default key deleted, `trust proxy` + right-to-left walk is the correct model. Crypto is right: fresh 12-byte nonce per assertion, version+timestamp bound as AAD, 32-byte key enforced, fail-closed when unkeyed. Wire format matches the verifier (context7app#1038).

Suggestions (non-blocking):
- `packages/mcp/src/index.ts:437` — `req.ip` can be IPv4-mapped (`::ffff:x.y.z.w`) on a direct socket and is signed verbatim. The verifier should unmap (flagged there); alternatively normalize here.
- `packages/mcp/src/index.ts:320` — `trust proxy ["loopback","linklocal","uniquelocal"]` excludes CGNAT (100.64/10). Confirm the ingress hop addresses fall in the trusted ranges, or the LB address gets signed.
- Avoid reusing the same key for the GCM assertion and the legacy CBC header in production (tests set both to one key); the rollout plan's dedicated secret covers this — just keep them distinct.

fahreddinozcan:
Review follow-up:

- Addressed CGNAT preservation in `e480063`: `100.64.0.0/10` is now explicitly included in the trusted proxy boundary.
- IPv4-mapped addresses are normalized by the verifier in context7app `015b174f`, preventing duplicate quota buckets.
- Production uses separate secret names/material for the GCM assertion and temporary CBC compatibility header. The shared deterministic test key is test-only.

PR #3092, opened 2026-08-27
feat(cli): add GitHub Copilot CLI setup target

fahreddinozcan:
## Summary

- add GitHub Copilot CLI detection and `--copilot` setup/remove flags
- support global `~/.copilot/mcp-config.json` and project `.mcp.json`, both using `mcpServers`
- write compatible HTTP and stdio entries with the documented `tools: ["*"]` default
- require explicit `--copilot` selection for project setup because `.mcp.json` is shared with Claude Code and cannot identify ownership safely
- preserve unrelated configuration during setup and removal

## Validation

- `pnpm --dir packages/cli typecheck`
- `pnpm --dir packages/cli lint:check`
- `pnpm --dir packages/cli test` — 325 tests passed
- `pnpm --dir packages/cli build`
- native `copilot mcp get` reports the generated entry as an enabled workspace server with all tools

Linear: [CTX7-2532](https://linear.app/upstash/issue/CTX7-2532/add-github-copilot-cli-as-a-ctx7-setup-target)

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -27,6 +27,7 @@ ctx7 setup --claude
 ctx7 setup --opencode
 ctx7 setup --vscode
 ctx7 setup --devin
+ctx7 setup --copilot
 ```
 
 ### Library Documentation
@@ -83,6 +84,7 @@ ctx7 setup --claude
 ctx7 setup --opencode
 ctx7 setup --vscode
 ctx7 setup --devin
+ctx7 setup --copilot
 
 # Use an existing API key instead of OAuth
 ctx7 setup --api-key YOUR_API_KEY
@@ -110,6 +112,7 @@ ctx7 remove --cursor
 ctx7 remove --claude --project
 ctx7 remove --vscode --project
 ctx7 remove --devin --project
+ctx7 remove --copilot --project
 
 # Remove both setup modes explicitly
 ctx7 remove --cursor --all
@@ -147,6 +150,7 @@ The CLI automatically detects which AI coding assistants you have installed and 
 | Cursor                                                              | `.cursor/skills/` |
 | VS Code                                                             | `.agents/skills/` |
 | Devin                                                               | `.devin/skills/`  |
+| GitHub Copilot CLI                                                  | `.agents/skills/` |
 | Antigravity                                                         | `.agent/skills/`  |
 
 ## Disabling Telemetry
--- a/packages/cli/src/__tests__/remove.test.ts
+++ b/packages/cli/src/__tests__/remove.test.ts
@@ -345,6 +345,35 @@ describe("remove command", () => {
     });
   });
 
+  test("removes Copilot CLI project config from .mcp.json", async () => {
+    const configPath = join(tempDir, ".mcp.json");
+
+    await writeFile(
+      configPath,
+      JSON.stringify(
+        {
+          mcpServers: {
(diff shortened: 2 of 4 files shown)


enesgules (changes requested):
The functional config is correct per the [Copilot CLI docs](https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-mcp-servers) (`~/.copilot/mcp-config.json` + `mcpServers`, project `.mcp.json`, `type: "stdio"` is a valid alias for local, skills in `.agents/skills` / `~/.agents/skills`). But two things should change before merge:

1. **Dead abstraction.** `projectConfigKey`, `buildProjectEntry`, and `getMcpConfigKey` are added and `scope` is threaded through `setup.ts` / `remove.ts`, but no agent sets the optional fields — both scopes resolve to `"mcpServers"` (the test even asserts this). This looks like a leftover from the abandoned `.vscode/mcp.json` design. Please delete it (~40 lines across 3 files) and keep using `agent.mcp.configKey`; reintroduce the indirection when an agent actually needs different keys per scope.

2. **Stale PR body.** The body says project `.vscode/mcp.json` with the `servers` key. The code uses `.mcp.json` + `mcpServers` — which is right, Copilot CLI does not read `.vscode/mcp.json`. Update the body.

Judgment calls to confirm:

- `tools: ["query-docs", "resolve-library-id"]` hard-codes the allowlist. If the MCP server adds or renames a tool, Copilot silently blocks it. `tools: ["*"]` (the documented default) avoids that — keep the explicit list only if intentional.
- Copilot shares `.mcp.json` + `mcpServers` with the `claude` agent, so `ctx7 remove --copilot --project` also removes Claude Code's project entry, and any repo with `.mcp.json` detects Copilot as installed. Probably fine since it's the same entry, but confirm it's intentional.

fahreddinozcan:
Addressed:

- removed `projectConfigKey`, `buildProjectEntry`, `getMcpConfigKey`, and the threaded scope indirection in 17efc39
- disabled ambiguous Copilot project auto-detection because `.mcp.json` is shared with Claude Code; explicit `--copilot --project` remains supported
- changed the tool policy to the documented `tools: ["*"]` default in 3e30d1e so future Context7 tools are not silently blocked
- corrected the PR description to `.mcp.json` + `mcpServers`

Explicit project removal still removes the shared `context7` entry by design: both clients consume that same named entry from the same file, so it cannot be removed for only one client. Auto-detection no longer attributes the shared file to Copilot.

Native `copilot mcp get` reports the generated config as an enabled workspace server with all tools. The full CLI suite, typecheck, formatting, lint, and build pass locally.

enesgules (approved):
All four points from my previous review are addressed:

1. The unused `projectConfigKey` / `buildProjectEntry` / `getMcpConfigKey` abstraction is gone — back to `agent.mcp.configKey`.
2. PR body now matches the code (`.mcp.json` + `mcpServers`).
3. `tools: ["*"]` replaces the hard-coded allowlist.
4. `detect.projectPaths` is now empty with a clear comment, so a Claude-only `.mcp.json` no longer auto-detects as Copilot; project setup stays available via explicit `--copilot`. There's a regression test for exactly this.

Remaining known caveat (fine to ship): `remove --copilot --project` and `remove --claude --project` operate on the same `.mcp.json` entry, since both tools read the same file and key — inherent to the shared format, not worth special-casing.

Approved.

PR #3093, opened 2026-08-27
fix(mcp): validate HTTP origins and hosts

fahreddinozcan:
## Summary

Fixes the MCP HTTP Origin validation / permissive CORS finding reported as Cobalt **PT38659_23**.

- rejects foreign, `null`, malformed, and explicitly empty Origin values before JSON parsing or MCP dispatch on loopback servers
- validates Host values on loopback servers to prevent DNS rebinding
- echoes approved loopback origins with complete `Vary` headers
- preserves wildcard CORS on non-loopback deployments because hosted authentication is header-only and browser MCP clients must remain compatible
- binds local HTTP to `127.0.0.1` by default and documents the exact client URL instead of relying on platform-dependent `localhost` resolution
- keeps the production container reachable with an explicit `0.0.0.0` bind
- scopes bind-host validation to HTTP mode, so `CONTEXT7_MCP_HOST` cannot break stdio startup
- adds focused unit and end-to-end regression coverage

Linear: [CTX7-2533](https://linear.app/upstash/issue/CTX7-2533/cwe-346cwe-942-validate-mcp-request-origins-and-restrict-local-http)

## Validation

- MCP suite: **110 passed**
- CLI suite: **361 passed**
- Pi integration suite: **4 passed**
- repository-wide format, typecheck, lint, and build passed
- latest-base merge check found no conflicts
- `git diff --check` passed


packages/mcp/src/lib/http-security.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    if (isLocal && !isLoopbackHost(req.headers.host)) {
+      res.status(403).json({ error: "forbidden", message: "Untrusted Host header." });
+      return;
+    }
+
+    const origin = req.headers.origin;
+    if (origin !== undefined && !isOriginAllowed(origin)) {
+      res.status(403).json({ error: "forbidden", message: "Untrusted Origin header." });
```
enesgules: **Blocking.** For a non-loopback bind this rejects every browser client that is not on `context7.com` or in `CONTEXT7_MCP_ALLOWED_ORIGINS`. The hosted server has no cookies, so the Origin allowlist gives no CSRF protection there; it only breaks MCP Inspector and other web-based MCP hosts against `mcp.context7.com`.

Suggest: keep `isLoopbackOrigin` when `isLocal`, and for the hosted branch allow any Origin (echo it, or send `*`). Keep the Host check as is. `DEFAULT_HOSTED_ORIGINS` and `normalizeConfiguredOrigins` can then go, along with the env var and its docs.

packages/mcp/src/lib/http-security.ts
```diff
@@ -0,0 +1,148 @@
...
+  bindHost: string,
+  additionalHostedOrigins?: string
+): RequestHandler {
+  const isLocal = isLoopbackHostname(bindHost);
+  const hostedOrigins = isLocal ? undefined : normalizeConfiguredOrigins(additionalHostedOrigins);
+  const isOriginAllowed = isLocal
+    ? isLoopbackOrigin
+    : (origin: string) => {
+        const parsed = parseOrigin(origin);
+        return parsed !== undefined && hostedOrigins?.has(parsed.origin) === true;
```
enesgules: Minor: `hostedOrigins` is always defined in this branch, so `?.` and `=== true` are noise. `hostedOrigins.has(parsed.origin)` is enough. (Moot if the blocking comment above is accepted.)

packages/mcp/src/index.ts
```diff
@@ -78,6 +85,13 @@ const CLI_PORT = (() => {
   return isNaN(parsed) ? undefined : parsed;
 })();
 
+let HTTP_HOST: string;
+try {
+  HTTP_HOST = normalizeBindHost(cliOptions.host);
```
enesgules: Minor: this runs in stdio mode too, so a bad `CONTEXT7_MCP_HOST` in the environment kills a stdio server. Move the call inside the `http` branch (or skip when `TRANSPORT_TYPE === "stdio"`).

packages/mcp/src/index.ts
```diff
@@ -538,7 +535,7 @@ async function main() {
     });
 
     const startServer = (port: number, maxAttempts = 10) => {
-      const httpServer = app.listen(port);
+      const httpServer = app.listen(port, HTTP_HOST);
```
enesgules: **Blocking.** With `127.0.0.1` only, clients that resolve `localhost` to `::1` and do not fall back fail with `ECONNREFUSED`. On macOS + Node 24, `dns.lookup("localhost", {all:true})` returns `::1` first. Node 20+ HTTP clients recover via `autoSelectFamily`, but older Node, Java, and some Go/Python clients try only the first address.

Options: (a) when `HTTP_HOST` is loopback, listen on both `127.0.0.1` and `::1`; or (b) keep single bind and change all `http://localhost:PORT/mcp` examples in README/docs to `http://127.0.0.1:PORT/mcp`. (a) is safer for existing user configs.

packages/mcp/README.md
```diff
@@ -1474,6 +1475,13 @@ You can use the `CONTEXT7_API_KEY` environment variable instead of passing the `
 CONTEXT7_API_KEY=your_api_key_here
 ```
 
+HTTP deployments can also set these security options:
+
+- `CONTEXT7_MCP_HOST` – Interface to bind; the CLI `--host` value takes precedence.
+- `CONTEXT7_MCP_ALLOWED_ORIGINS` – Comma-separated additional browser origins allowed to call the server. Values must be exact origins, such as `https://docs.example.com`.
+
+Requests without an `Origin` header, such as normal server-to-server MCP clients, continue to work. When bound to loopback, browser requests must use a loopback origin and the Host header is validated against DNS rebinding. With a non-loopback bind, Context7's production web origins and any explicitly configured origins are allowed.
```
enesgules: If the hosted-mode change is accepted, update this sentence and the matching text in `docs/resources/developer.mdx`: with a non-loopback bind all origins are allowed and only the Host check applies to loopback.

enesgules (changes requested):
The loopback logic is correct and well tested: IPv6 brackets, `127.1` canonicalization, userinfo/path rejection, `null` origin rejection, rejection before `express.json`, `Vary` headers, `getOptionValueSource`. Two blocking items on scope and compatibility, plus a few minor ones inline.

**Blocking**

1. **Hosted mode breaks browser MCP clients.** The public server at `mcp.context7.com` has no cookies and no ambient credentials; auth is header-only. For that server `Access-Control-Allow-Origin: *` is not a vulnerability. The comment this PR removes from `index.ts` shows the team deliberately added those headers so browser clients pass preflight. After this change MCP Inspector, web IDEs, and the Cloudflare playground get 403 from the hosted server. The Cobalt finding is the local server + DNS rebinding, and that stays fixed if the strict Origin allowlist applies only when `isLocal` is true. See inline comment.

2. **`localhost` clients can get `ECONNREFUSED`.** The default bind is now `127.0.0.1` only. On macOS with Node 24, `localhost` resolves to `::1` first (verified locally: a Node client with `autoSelectFamily: false` against a `127.0.0.1`-only listener fails; `autoSelectFamily: true` succeeds). Every doc and MCP config example says `http://localhost:3000/mcp`. Either bind both `127.0.0.1` and `::1` for a loopback host, or change the docs/README examples to `http://127.0.0.1:PORT/mcp`. See inline comment.

packages/mcp/src/lib/http-security.ts
fahreddinozcan: Addressed in 34a007c. Non-loopback deployments now preserve wildcard CORS, while strict Origin and Host validation remains scoped to loopback servers. The hosted allowlist, environment variable, parsing branch, and docs were removed.

packages/mcp/src/index.ts
fahreddinozcan: Addressed in 34a007c using option (b). The server keeps the single 127.0.0.1 bind, and both MCP docs now give the exact 127.0.0.1 client URL and explicitly warn against substituting localhost because its resolution is platform-dependent.

packages/mcp/src/index.ts
fahreddinozcan: Addressed in 34a007c. Bind-host normalization now runs only inside the HTTP branch, with an integration test proving that an invalid CONTEXT7_MCP_HOST value does not break stdio mode.

packages/mcp/README.md
fahreddinozcan: Addressed in 34a007c. Both docs now describe strict loopback validation and wildcard CORS for non-loopback deployments; the hosted-origin environment variable was removed.

packages/mcp/src/lib/http-security.ts
fahreddinozcan: Addressed in 34a007c by removing the hosted-origin set and its optional lookup entirely.

enesgules (approved):
All four threads are addressed in 34a007c and verified locally: hosted mode keeps wildcard CORS, strict Origin/Host checks are loopback-only, bind-host normalization runs only in the HTTP branch, docs use 127.0.0.1, and Dockerfile/Smithery keep 0.0.0.0. Typecheck, lint, build, and 94 tests pass.

PR #3101, opened 2026-08-28
fix(mcp): increase HTTP subscription capacity

fahreddinozcan:
## Summary

- raise the HTTP `subscriptions/listen` capacity from the SDK default of 1,024 to 16,000 per process
- allow operators to override it with `MCP_MAX_SUBSCRIPTIONS`
- validate overrides, warn on invalid values, and safely fall back to 16,000
- leave stdio transport and existing HTTP keepalive behavior unchanged

## Why

Production MCP pods are healthy at the Envoy layer, but every pod is repeatedly logging `subscription limit reached (1024)`. The SDK reports that rejection as an in-band JSON-RPC error over HTTP 200, which is why connection/request overflow and 5xx metrics remain zero.

The selected default is deliberately conservative: 16,384 streams stayed near baseline request latency in the live-Docker benchmark, while 32,768 raised `tools/list` p95 to about 39 ms. A 16,000 default preserves the capacity guard, provides about 15.6x the SDK default per pod, and allows 160,000 streams across the current 10-pod deployment.

## Docker capacity benchmark

Exact PR image: Node LTS Alpine, arm64 Docker VM with 12 CPUs and 7.65 GiB memory, no container CPU/memory limit, nofile 1,048,576. Server and load generator used an isolated Docker network.

| Open streams | Server memory | Idle CPU | `tools/list` p95 | Result |
| ---: | ---: | ---: | ---: | --- |
| 8,192 | 706 MiB | 0.02% | 3.30 ms | stable |
| 16,384 | 1.04 GiB | 0.00% | 3.37 ms | stable |
| 24,576 | 1.51 GiB | 0.00% | 3.27 ms | stable |
| 32,768 | 1.59 GiB | 0.00% | 38.68 ms steady-state | rejected as default |

At 24,576 streams, an additional 2,000 `tools/list` requests at concurrency 50 completed with 0 errors at 1,745 req/s. The zero-stream control completed at 1,685 req/s in the same setup.

Production MCP workers are m8g.2xlarge nodes with about 30 GiB allocatable memory, two MCP pods per node, and current node memory use around 11–12%, so the tested capacity fits with substantial headroom.

## Verification

- `pnpm --filter @upstash/context7-mcp format:check`
- `pnpm --filter @upstash/context7-mcp test` (59/59 passed; HTTP modern/legacy and stdio modern/legacy)
- `pnpm --filter @upstash/context7-mcp build`
- `pnpm --filter @upstash/context7-mcp lint:check`
- live Docker subscription and concurrent-request benchmarks described above

### DIFF
--- a/.changeset/quiet-dodos-listen.md
+++ b/.changeset/quiet-dodos-listen.md
@@ -1,0 +1,5 @@
+---
+"@upstash/context7-mcp": patch
+---
+
+Increase the default HTTP subscription capacity and allow deployments to configure it with `MCP_MAX_SUBSCRIPTIONS`.
--- a/.env.example
+++ b/.env.example
@@ -13,6 +13,9 @@ OPENAI_APPS_CHALLENGE_TOKEN=
 # Required for hosted HTTP deployments. Must match context7.com.
 MCP_CLIENT_IP_ASSERTION_KEY=
 
+# Maximum concurrent MCP subscription streams per HTTP process.
+MCP_MAX_SUBSCRIPTIONS=4096
+
 # Temporary legacy rollout key. Removal is tracked by CTX7-2536.
 CLIENT_IP_ENCRYPTION_KEY=
 
--- a/.env.example
+++ b/.env.example
@@ -14,7 +14,7 @@ OPENAI_APPS_CHALLENGE_TOKEN=
 MCP_CLIENT_IP_ASSERTION_KEY=
 
 # Maximum concurrent MCP subscription streams per HTTP process.
-MCP_MAX_SUBSCRIPTIONS=24000
+MCP_MAX_SUBSCRIPTIONS=16000
 
 # Temporary legacy rollout key. Removal is tracked by CTX7-2536.
 CLIENT_IP_ENCRYPTION_KEY=
--- a/.env.example
+++ b/.env.example
@@ -14,7 +14,7 @@ OPENAI_APPS_CHALLENGE_TOKEN=
 MCP_CLIENT_IP_ASSERTION_KEY=
 
 # Maximum concurrent MCP subscription streams per HTTP process.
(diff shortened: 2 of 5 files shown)


enesgules (approved):
Approving — the change is mechanically correct: `maxSubscriptions` is a real option on `createMcpHandler` in `@modelcontextprotocol/server` 2.0.0 (SDK default 1024), the handler is created once per process so the limit is per-process as described, parsing edge cases (`""`, `"Infinity"`, negatives, floats) all fall back correctly, and CI passes.

Notes to address:

1. **Default value mismatch (please fix before merge).** The PR body says the new default is 24,000 and the benchmark plateau is 24,576, but the code ships `DEFAULT_MAX_SUBSCRIPTIONS = 16_000` and `.env.example` shows 16000. If 16,000 is the deliberate final choice, update the PR body and the "240,000 streams across 10 pods" claim (it becomes 160,000); otherwise update the code.

2. **Silent fallback.** An invalid `MCP_MAX_SUBSCRIPTIONS` value returns the default with no signal. A single `console.warn` on the invalid path would help operators catch typos (the repo already does this pattern in `lib/encryption.ts`).

3. **No upper bound.** The benchmark rejected 32,768 because tools/list p95 degraded from ~3 ms to ~39 ms, but the override accepts any positive integer. No cap needed, but a comment on the constant recording the 32k finding would preserve that knowledge.

fahreddinozcan:
Addressed the review notes in aa9201d: the PR description now consistently documents the 16,000-per-process default and 160,000 total across 10 pods; invalid MCP_MAX_SUBSCRIPTIONS values emit an operator warning before falling back; and the source records the 32,768-stream latency finding without imposing a new hard cap. Focused tests, typecheck, lint, and format checks pass.

PR #3109, opened 2026-08-30
fix(cli): reuse device login API key during setup

### LINKED ISSUE #3107: [Bug]: ctx7 setup (MCP mode) fails with "Please sign in to create API keys": device flow returns an API key, setup sends it to dashboard endpoint
**Version:** ctx7 0.5.9 (npx), matches master

**Summary**

`ctx7 login` succeeds, but `ctx7 setup` (MCP mode) always fails: `✖ Authentication failed / ✖ Please sign in to create API keys`.

**Root cause**

The device-flow token endpoint now returns a long-lived API key (`access_token: "ctx7sk-..."`, `scope: "api-key"` in `~/.config/context7/credentials.json`). `setup` in MCP mode sends that key as a Bearer token to `POST /api/dashboard/api-keys` to create a *new* key (`packages/cli/src/commands/setup.ts`, `authenticateAndGenerateKey()`). Dashboard endpoints require a session/OAuth token, not a `ctx7sk-` key, so it 401s.

**Repro**

1. `npx ctx7 login` → approve in browser → "Login successful!" (note: generic message, because the internal `whoami` call also 401s)

2. `npx ctx7 setup --opencode` → choose MCP server → fails as above
3. Repeat `logout` + `login`; same result every time

**Evidence**

With the stored key (valid for docs/MCP, both return 200):

- `GET /api/dashboard/whoami` → 401 `{"message":"Unauthorized"}`
- `GET /api/dashboard/api-keys` → 401 `{"message":"Please sign in to list API keys"}`

**Expected**

If the stored token is already an API key, `setup` should use it directly (the CLI+Skills mode path, `resolveCliAuth`, already does this and works: `setup --opencode --cli` succeeds).

**Workarounds**

`setup --cli`, or `setup --mcp --api-key <stored ctx7sk key>`, or `--oauth`.

fahreddinozcan:
## Summary

Fixes #3107.

The OAuth device flow returns a long-lived Context7 API key (`ctx7sk-…`) as its access token. MCP setup was treating that credential as a dashboard session token and sending it to `POST /api/dashboard/api-keys` to mint another key. Once Context7app restricted dashboard routes to Clerk session/OAuth credentials, setup began failing with “Please sign in to create API keys.”

This change:

- centralizes Context7 API-key recognition in the CLI auth utility
- uses device-flow API keys directly during MCP setup
- keeps the legacy OAuth-to-API-key exchange path for older credentials
- isolates setup credential resolution in a focused module
- prevents API-key login and `whoami` from probing the session-only dashboard identity endpoint
- updates the CLI documentation and bundled setup reference to describe the actual device-flow contract

## Rollout

Context7app PR #1057 temporarily restores the old dashboard authentication behavior for released CLI versions. After this fix is published and users have an upgrade window, that rollback should be reversed so dashboard endpoints are session/OAuth-only again.

## Validation

- `pnpm --filter ctx7 test` — 320 passed
- `pnpm --filter ctx7 format:check`
- `pnpm --filter ctx7 build`
- `git diff --check`
- Manual device-key setup succeeded with the dashboard base deliberately unreachable and wrote the stored API key to MCP configuration
- Manual legacy-OAuth setup made the expected authenticated dashboard key-generation request and wrote the returned API key to MCP configuration
- Manual API-key `whoami` succeeded against an unreachable dashboard base, printed only `Logged in`, and made no network request

### DIFF
--- a/.changeset/fuzzy-ravens-setup.md
+++ b/.changeset/fuzzy-ravens-setup.md
@@ -1,0 +1,5 @@
+---
+"ctx7": patch
+---
+
+Use API keys returned by the device login flow directly during MCP setup instead of sending them to a dashboard session endpoint to generate another key.
--- a/docs/clients/cli.mdx
+++ b/docs/clients/cli.mdx
@@ -148,7 +148,7 @@ ctx7 setup --api-key YOUR_API_KEY
 ctx7 setup --oauth
 ```
 
-Without `--api-key` or `--oauth`, setup runs the OAuth device flow: it shows a verification link and short code that you open on any device to sign in, so it works the same locally or on a remote, headless, or SSH host. MCP mode additionally generates a new API key after login. `--oauth` is MCP-only — use it when an IDE handles the auth flow on your behalf.
+Without `--api-key` or `--oauth`, setup runs the OAuth device flow: it shows a verification link and short code that you open on any device to sign in, so it works the same locally or on a remote, headless, or SSH host. The device flow returns an API key that setup uses for authentication. `--oauth` is MCP-only — use it when an IDE handles the auth flow on your behalf.
 
 **What gets written — MCP mode:**
 
--- a/packages/cli/src/__tests__/auth-commands.test.ts
+++ b/packages/cli/src/__tests__/auth-commands.test.ts
@@ -7,14 +7,17 @@ const mockSaveTokens = vi.fn();
 const mockStartDeviceAuthorization = vi.fn();
 const mockPollDeviceToken = vi.fn();
 
-vi.mock("../utils/auth.js", () => ({
-  getValidAccessToken: (...args: unknown[]) => mockGetValidAccessToken(...args),
-  clearTokens: (...args: unknown[]) => mockClearTokens(...args),
-  saveTokens: (...args: unknown[]) => mockSaveTokens(...args),
-  startDeviceAuthorization: (...args: unknown[]) => mockStartDeviceAuthorization(...args),
-  pollDeviceToken: (...args: unknown[]) => mockPollDeviceToken(...args),
-  DEFAULT_DEVICE_POLL_INTERVAL_SECONDS: 5,
-}));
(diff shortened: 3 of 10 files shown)


enesgules (approved):
The fix is correct and finds the root cause from #3107. The `isContext7ApiKey()` guard stops the bad key exchange, and the legacy OAuth path stays for old credentials. The extracted `resolveSetupApiKey` is a line-for-line move of the old `authenticateAndGenerateKey` with only the guard added. Ran `pnpm --filter ctx7 test` (319 passed) and `typecheck` locally on the branch.

Minor points, not blockers:

1. **Same root cause remains in `whoami`.** `commands/auth.ts` still sends the `ctx7sk-` key to `/api/dashboard/whoami`. When context7app reverses the rollback, `ctx7 whoami` will show "Session may be expired" and login will show the generic "Login successful!" message (issue #3107 reports this symptom). Needs a tracked follow-up: accept API keys on `whoami` server-side, or add a key-compatible identity endpoint.
2. **Duplicated prefix in the test mock.** `setup-auth.test.ts` re-implements `startsWith("ctx7sk-")` instead of using the real `isContext7ApiKey`. If the prefix changes, this test stays green while setup breaks.
3. **New module for one function.** The guard alone inside `setup.ts` would have been a sufficient diff; the extraction is acceptable for testability but adds a directory for one export.

fahreddinozcan:
Addressed the review notes in `daa7a20`:

- setup/auth tests now retain the real `isContext7ApiKey` implementation via partial module mocks, so prefix drift cannot leave the orchestration tests falsely green
- API-key login identity and `ctx7 whoami` now skip `/api/dashboard/whoami`; `whoami` reports `Authentication: API key` instead of suggesting that a valid key is an expired session
- kept the focused `setup/auth.ts` module because it removes the credential exchange from the broad setup command and provides an intentional test boundary

Revalidated with 320 tests, typecheck, lint, formatting, build, diff check, and a manual API-key `whoami` run against an unreachable dashboard base.

PR #3112, opened 2026-08-31
fix(mcp): remove legacy client IP encryption

fahreddinozcan:
## Summary

- remove the transitional AES-CBC `mcp-client-ip` header
- remove MCP-side `CLIENT_IP_ENCRYPTION_KEY` configuration and test setup
- keep the authenticated AES-256-GCM assertion path unchanged
- correct the client-IP encryption description in the privacy documentation

This completes the remaining legacy cleanup for Cobalt PT38659_25 and tracks [CTX7-2580](https://linear.app/upstash/issue/CTX7-2580/harden-mcp-forwarded-client-ip-handling-and-remove-legacy-encryption).

## Validation

- `pnpm typecheck`
- `pnpm lint:check`
- `pnpm test` — 6 files, 59 tests
- `pnpm exec prettier --check ...`
- `git diff --check`

The integration suite builds and runs the real HTTP and stdio servers. It confirms that HTTP requests emit only the authenticated assertion, attacker-controlled forwarded-header prefixes are not selected, and stdio emits no client-IP assertion.

## Deployment

Remove `CLIENT_IP_ENCRYPTION_KEY` from the MCP deployment after rollout. Do not remove or rotate the context7app value used for OAuth token encryption.

Proxy handling and stdio behavior are unchanged.

### DIFF
--- a/.changeset/remove-legacy-client-ip.md
+++ b/.changeset/remove-legacy-client-ip.md
@@ -1,0 +1,5 @@
+---
+"@upstash/context7-mcp": patch
+---
+
+Remove the legacy AES-CBC client-IP header now that authenticated assertions are deployed.
--- a/.env.example
+++ b/.env.example
@@ -16,9 +16,6 @@ MCP_CLIENT_IP_ASSERTION_KEY=
 # Maximum concurrent MCP subscription streams per HTTP process.
 MCP_MAX_SUBSCRIPTIONS=16000
 
-# Temporary legacy rollout key. Removal is tracked by CTX7-2536.
-CLIENT_IP_ENCRYPTION_KEY=
-
 # Network / certificates
 HTTPS_PROXY=
 NODE_EXTRA_CA_CERTS=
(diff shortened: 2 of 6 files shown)


enesgules (approved):
LGTM — clean removal of the transitional AES-CBC `mcp-client-ip` path. Verified the GCM assertion path is untouched, fail-closed behavior is preserved and tested, no dangling references to `CLIENT_IP_ENCRYPTION_KEY` or `mcp-client-ip` remain, and the docs wording now matches the code.

Two non-blocking notes:

1. Safety of this release assumes the consumer-side rollout (CTX7-2580) is 100% complete — any un-patched API deployment still reading only `mcp-client-ip` would silently lose client IPs for rate limiting. Please confirm before merge.
2. The removed comments referenced CTX7-2536; make sure that ticket gets closed alongside CTX7-2580.

Keep the deployment note (remove `CLIENT_IP_ENCRYPTION_KEY` from the MCP deployment only; don't touch the context7app OAuth key) with the merge.

PR #3137, opened 2026-09-03
feat(sdk): add production HTTP controls and fix review findings

fahreddinozcan:
## Summary
- fix SDK response inference for runtime-selected formats and preserve optional/undefined option forwarding
- honor `retry: false` as exactly one request
- add production HTTP controls while keeping API-key authentication required
- align shared HTTP conventions with the latest `@upstash/redis` SDK
- decompose transport types, cancellation, retry policy, response decoding, and orchestration
- split deterministic unit tests from live API integration tests

## Production SDK controls
- client and per-request timeouts and abort signals, including a fresh signal factory per request
- configurable transient retries with the `@upstash/redis` retry shape and default backoff
- `Retry-After` support and GET-only retries; mutating requests remain single-attempt
- configurable `keepAlive`, custom `fetch`, `baseUrl`, additional headers, and native fetch cache settings
- HTTP(S) URL validation and typed JSON parse failures with truncated bodies and causes
- response metadata hooks for request IDs, rate limits, status, and attempt count
- structured `Context7Error` fields including status, code, request ID, rate limits, retryability, and cause
- protected `Authorization` header sourced from the required API key

## Maintainability
- reduce `http/index.ts` from 521 lines to 184 focused orchestration lines
- isolate cancellation, response parsing, retry policy, and public transport types
- use the canonical request contract in commands instead of duplicating transport fields
- fix timeout classification when a deadline expires during retry backoff
- remove speculative retry-method configuration from the public API

## Verification
- `pnpm --filter @upstash/context7-sdk test` (36 tests)
- `pnpm --filter @upstash/context7-sdk test:integration`
- `pnpm --filter @upstash/context7-sdk typecheck`
- `pnpm --filter @upstash/context7-sdk lint:check`
- `pnpm --filter @upstash/context7-sdk build`
- `pnpm --filter @upstash/context7-tools-ai-sdk typecheck`
- `pnpm --filter @upstash/context7-tools-ai-sdk build`

Authentication remains mandatory. Context7 intentionally extends the Redis transport convention by retrying selected transient HTTP statuses, honoring `Retry-After`, and restricting automatic retries to safe requests.


packages/sdk/src/client.ts
```diff
@@ -50,7 +50,7 @@ export class Context7 {
   async searchLibrary(
     query: string,
     libraryName: string,
-    options: SearchLibraryOptions & { type: "json" }
+    options: SearchLibraryOptions & { type?: "json" }
```
enesgules: Make this optional so it also covers the two-argument call, then delete the separate two-argument overload below:

```suggestion
    options?: SearchLibraryOptions & { type?: "json" }
```

packages/sdk/src/client.ts
```diff
@@ -65,11 +65,16 @@ export class Context7 {
...
+
+  /**
+   * Search for libraries with options whose response type is determined at runtime
+   */
   async searchLibrary(
     query: string,
     libraryName: string,
-    options?: SearchLibraryOptions
-  ): Promise<Library[]>;
+    options: SearchLibraryOptions
```
enesgules: Must be optional, otherwise `searchLibrary(q, n, undefined)` and `searchLibrary(q, n, opts as SearchLibraryOptions | undefined)` fail with TS2769 (they compiled before this PR):

```suggestion
    options?: SearchLibraryOptions
```

packages/sdk/src/client.ts
```diff
@@ -93,7 +98,7 @@ export class Context7 {
   async getContext(
     query: string,
     libraryId: string,
-    options: GetContextOptions & { type: "json" }
+    options: GetContextOptions & { type?: "json" }
```
enesgules: Same as `searchLibrary`: make optional and drop the two-argument overload.

```suggestion
    options?: GetContextOptions & { type?: "json" }
```

packages/sdk/src/client.ts
```diff
@@ -108,11 +113,16 @@ export class Context7 {
...
+
+  /**
+   * Get documentation context with options whose response type is determined at runtime
+   */
   async getContext(
     query: string,
     libraryId: string,
-    options?: GetContextOptions
-  ): Promise<Documentation[]>;
+    options: GetContextOptions
```
enesgules: Same regression as `searchLibrary`; `undefined` must stay accepted here.

```suggestion
    options?: GetContextOptions
```

packages/sdk/src/http/index.ts
```diff
@@ -112,7 +112,7 @@ export class HttpClient implements Requester {
     this.retry =
       typeof config?.retry === "boolean" && config?.retry === false
         ? {
-            attempts: 1,
+            attempts: 0,
```
enesgules: Nit, since this line is already touched: the condition two lines up is just `config.retry === false`.

packages/sdk/src/client.integration.test.ts
```diff
@@ -0,0 +1,120 @@
+import { describe, test, expect } from "vitest";
+import { Context7 } from "./client";
+
+describe("Context7 Client integration", () => {
+  const apiKey = process.env.CONTEXT7_API_KEY || process.env.API_KEY!;
```
enesgules: Nit: `API_KEY` is a dead fallback. The client only reads `CONTEXT7_API_KEY`.

packages/sdk/src/client.types.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  type SearchLibraryOptions,
+} from "./client";
+
+function searchWithOptions(client: Context7, options: SearchLibraryOptions) {
+  return client.searchLibrary("query", "react", options);
+}
+
+function getContextWithOptions(client: Context7, options: GetContextOptions) {
+  return client.getContext("query", "/react/react", options);
+}
```
enesgules, line 16: Please add a case with `options: SearchLibraryOptions | undefined` (and the same for `getContext`). That is the shape that currently fails to compile.

enesgules (changes requested):
Requesting changes for one type regression. The rest looks good.

**Must fix: the new overloads reject `undefined` options.** Overloads 1 and 4 now require `options`, and overload 3 takes only two arguments. A caller that forwards an optional value no longer compiles (verified with `tsc --strict`):

```ts
const o: SearchLibraryOptions | undefined = ...;
client.searchLibrary("q", "react", o);          // TS2769: No overload matches this call
client.searchLibrary("q", "react", undefined);  // TS2769
```

Both compiled before this PR. Suggested fix (inline comments below): make `options` optional in the JSON overload and the runtime-union overload, and drop the two-argument overload. Three overloads instead of four, and all call shapes compile. Same for `getContext`. Please add a type test for the `Options | undefined` case.

**Flag: changeset is `patch`, but this is a breaking type change.** Consumers that pass runtime-selected options and use the result as an array will fail to compile after upgrading. The new type is correct, but say so in the changeset text so users know why their build broke. At 0.x, `patch` is acceptable.

**Verified correct:**
- `retry: false` → `attempts: 0` is exactly one fetch with the `i <= attempts` loop. The new test proves it.
- Unit/integration split is clean; fork PRs no longer need the API secret. CI is green.
- `requesterWith` is the right size for a test double.

**Nits (optional):**
- The empty-input rejection is tested twice (`client.test.ts` and `search-library/index.test.ts`). Keep one.

packages/sdk/src/http/response.ts
```diff
@@ -0,0 +1,139 @@
...
+  response: Response,
+  metadata: Context7ResponseMetadata,
+  retryable: boolean
+): Promise<never> {
+  const rawBody = await response.text();
+  let errorBody: { error?: string; message?: string } = {};
+
+  if (rawBody) {
+    try {
+      errorBody = JSON.parse(rawBody) as typeof errorBody;
```
enesgules: `JSON.parse` can return `null`, an array, or a primitive here, and line 71 then reads `.message` on it. Keep the object shape only:

```suggestion
      const parsed: unknown = JSON.parse(rawBody);
      if (typeof parsed === "object" && parsed !== null) errorBody = parsed as typeof errorBody;
```

enesgules (changes requested):
All seven earlier comments are addressed in 097d280 (optional `options` overloads, the `| undefined` type test, `retry === false` in one place, dead `API_KEY` fallback removed). I resolved those threads.

I also reviewed the four feature commits added after the first review (HTTP controls, redis-js alignment, transport split, retry API). Typecheck, lint, build, 36 unit tests, and the ai-sdk typecheck pass locally. The design is sound: GET-only retries, `Retry-After`, timeout classification during backoff, protected `Authorization`, and the abort-state cleanup all check out.

**Must fix (one line):** an error response whose body is the JSON literal `null` throws a raw `TypeError` instead of a `Context7Error`. Repro:

```ts
const client = new HttpClient({ baseUrl: "https://x.test", retry: false,
  fetch: async () => new Response("null", { status: 500, headers: { "content-type": "application/json" } }) });
await client.request({ method: "GET", path: ["a"] });
// TypeError: Cannot read properties of null (reading 'message')
```

Suggestion inline. A unit test next to "falls back to statusText on empty error body" would lock it in.

**Non-blocking**
- The changeset says timeouts are configurable but not that a 30 s default now applies to every request. Master had no timeout. One sentence in the changeset would help users who see new `request_timeout` errors.
- `Command.requestResult` now throws `TypeError` for a missing result where master threw `Context7Error`. The path is unreachable today, but `Context7Error` keeps the public contract consistent.

packages/sdk/src/http/response.ts
fahreddinozcan: ctx7-2663: Fixed in 813bb6a. Parsed error bodies are now shape-checked before accessing fields, and the regression test covers the JSON literal null.

enesgules (approved):
813bb6a addresses everything: non-object JSON error bodies are shape-checked with a regression test for the null literal, Command.requestResult throws Context7Error again, and the changeset documents the 30 s default timeout. Verified locally: typecheck, lint, build, ai-sdk typecheck, and 38 unit tests pass. CI is green.

PR #3145, opened 2026-09-04
CTX7-2692: Support Vercel Marketplace OIDC in SDK and MCP

fahreddinozcan:
## Summary

- add an SDK `authToken` value/provider alongside API-key authentication
- resolve async token providers once per Context7 request so short-lived Marketplace credentials are refreshed instead of cached for the client lifetime
- validate Vercel Marketplace tokens at the MCP edge using the configured integration issuer JWKS, exact product audience, RS256, `nbf`/`exp`, and immutable `resource` claim
- document per-resource bearer-token use without depending on the not-yet-released Vercel token-minting API signature

## Compatibility

- explicit credentials take precedence as `apiKey`, `authToken`, then `CONTEXT7_API_KEY`, allowing Vercel's Phase 1 dual-auth rollout to exercise OIDC while the legacy secret remains
- custom `Authorization` headers remain protected from override
- no production data or credentials are changed

## Verification

- SDK: 40 tests (including a legacy environment key), typecheck, ESLint, and tsup build
- MCP Marketplace auth: 17 tests, typecheck, ESLint, and TypeScript build
- `git diff --check`

## Rollout dependency

The final Vercel Marketplace token-minting API is not yet available. The SDK accepts its short-lived bearer tokens through a stable provider callback; the README can be updated to the final Vercel helper signature once Vercel publishes it.

Linear: https://linear.app/upstash/issue/CTX7-2692/oidc-support

### DIFF
--- a/.changeset/soft-dodos-oidc.md
+++ b/.changeset/soft-dodos-oidc.md
@@ -1,0 +1,6 @@
+---
+"@upstash/context7-sdk": minor
+"@upstash/context7-mcp": minor
+---
+
+Add per-request OIDC bearer token support for Vercel Marketplace connections.
--- a/.changeset/soft-dodos-oidc.md
+++ b/.changeset/soft-dodos-oidc.md
@@ -3,4 +3,4 @@
 "@upstash/context7-mcp": minor
 ---
 
-Add per-request OIDC bearer token support for Vercel Marketplace connections.
+Add per-request bearer-token providers and Vercel Marketplace resource OIDC validation.
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,10 @@ EMA_ISSUER=
 AUTH_SERVER_URL=
 OPENAI_APPS_CHALLENGE_TOKEN=
 
+# Exact values assigned to the Context7 Vercel Marketplace product.
+VERCEL_MARKETPLACE_OIDC_ISSUER=
+VERCEL_MARKETPLACE_OIDC_AUDIENCE=
+
 # Required for hosted HTTP deployments. Must match context7.com.
 MCP_CLIENT_IP_ASSERTION_KEY=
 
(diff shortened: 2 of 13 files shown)


enesgules (changes requested):
Validation logic matches the Vercel "Marketplace OIDC for Providers" doc (issuer, audience, JWKS path, clock tolerance, `resource` claim). SDK per-request token provider is the right shape for short-lived tokens. One README regression to fix before merge, rest are nits.

Blocking:

- `packages/mcp/README.md`: the change collapses the alert into `> [!TIP] > **Recommended Post-Setup...**` on one line. GitHub only renders the alert when `[!TIP]` is alone on the first line, so the tip box breaks in the most-viewed README we have. Restore the two lines.

Non-blocking:

- `packages/mcp/README.md`: the Vercel section is inserted in the middle of Requirements, above the tip. Move it to the bottom of the README (or docs site). Most readers of this file are not Marketplace users.
- Both READMEs declare a made-up `getMarketplaceResourceToken()`. The Vercel doc names the real helper: `import { getIntegrationToken } from '@vercel/integrations'`. Use `authToken: () => getIntegrationToken('context7')` and note the signature is subject to change.
- `packages/sdk/src/http/index.ts`: the token is resolved once before `fetchWithRetry`. A retry after a 429 backoff can resend an expired token. Resolve inside each attempt, or accept and document.
- `packages/mcp/src/lib/vercelMarketplaceJwt.ts`: `algorithms: ["RS256"]` is not in the Vercel doc. Vercel's current marketplace JWKS is RS256, so fine today; confirm with Vercel once test tokens exist.
- `packages/sdk/src/client.ts`: the `Credential` union only feeds the `ctx7sk` prefix warning. Optional simplification.

fahreddinozcan:
Addressed the review feedback in `08ada60`:

- Restored the GitHub TIP alert formatting and moved the Vercel OIDC section to the bottom of the MCP README.
- Replaced the placeholder helper with `getIntegrationToken` from `@vercel/integrations`, including the pre-GA signature note.
- The SDK now resolves a fresh auth token for every retry attempt, covered by a regression test.
- Kept RS256 pinned for fail-closed behavior with the current Vercel JWKS. We will confirm this once Vercel provides test tokens.
- Retained the credential union because it preserves the useful static Context7-key warning while callback tokens remain opaque.

Local verification passed: SDK 42/42 tests, MCP 67/67 tests, both typechecks, both lints, and both builds. @enesgules please re-review when convenient.

enesgules (approved):
README alert restored, Vercel section moved to the bottom, examples use `getIntegrationToken` from `@vercel/integrations`, and the token now resolves per retry attempt with a test. CI green. Approving.

PR #3166, opened 2026-09-09
docs: add Cursor Cloud Agents setup guide

fahreddinozcan:
## Summary

- add a step-by-step guide for authenticated Context7 access in Cursor Cloud Agents
- document the verified custom HTTP MCP header configuration and OAuth plugin behavior
- add six redacted, tightly cropped HD UI screenshots plus verification and troubleshooting steps
- keep the common Cursor installation and usage path ahead of the Cloud-specific guide

## Verification

- reproduced Context7 MCP calls from a Cursor Cloud Agent
- verified API-key authentication with the Context7-API-Key header
- checked MDX formatting with Prettier
- checked the diff for whitespace errors and confirmed no API key was included


docs/clients/cursor.mdx
```diff
@@ -19,6 +19,148 @@ For manual MCP installation or other configuration options, see [All MCP Clients
...
+
+<Step title="Configure the authenticated Context7 server">
+Enter the following values:
+
+| Field | Value |
+| --- | --- |
+| **Name** | `context7-api-key` |
+| **Type** | `URL` |
+| **Server URL** | `https://mcp.context7.com/mcp` |
+| **Header key** | `Context7-API-Key` |
```
enesgules, line 133: Every other doc (all-clients intro, Claude Code, the existing Cursor accordion) uses `Authorization: Bearer YOUR_API_KEY`. The server accepts both, but introducing a second header name here with no explanation will confuse readers. Use `Authorization` / `Bearer YOUR_API_KEY` in the table and JSON below to match, or add one line saying why this guide differs.

docs/clients/cursor.mdx
```diff
@@ -19,6 +19,148 @@ For manual MCP installation or other configuration options, see [All MCP Clients
...
+| **Header value** | Your Context7 API key |
+
+<Warning>
```
enesgules, line 140: This is factually wrong. The server reads `context7_api_key` as a header too (see `extractApiKey` in `packages/mcp/src/index.ts`). The likely real cause is that Cursor's backend proxy drops header names containing underscores, which is standard nginx behavior. Suggest: "Use a hyphenated header name. Header names with underscores can be dropped by Cursor's proxy."

docs/clients/cursor.mdx
```diff
@@ -19,6 +19,148 @@ For manual MCP installation or other configuration options, see [All MCP Clients
...
+  <img src="/images/clients/cursor/cloud-agents/03-add-custom-mcp.jpg" alt="Custom MCP option in Cursor's MCP browser" />
+</Frame>
+</Step>
+
+<Step title="Configure the authenticated Context7 server">
+Enter the following values:
+
+| Field | Value |
+| --- | --- |
+| **Name** | `context7-api-key` |
```
enesgules, line 130: `context7-api-key` leaks the auth mechanism into the tool namespace. `context7` matches the other docs. If the distinct name is intentional (to avoid a collision with the installed plugin), say so.

docs/clients/cursor.mdx
```diff
@@ -19,6 +19,148 @@ For manual MCP installation or other configuration options, see [All MCP Clients
...
+</Step>
+</Steps>
+
+### OAuth and the Context7 Cursor plugin
+
+The Context7 plugin uses the OAuth endpoint and does not currently provide a field for an API key. Cursor supports OAuth for Cloud MCP servers, but installing or enabling the plugin does not guarantee that its OAuth sign-in flow has completed.
+
+- Use the plugin when you want OAuth and complete the sign-in prompt for the same Cursor account that launches the Cloud Agent.
+- Use the custom HTTP setup above when you need explicit API-key authentication, including access to private Context7 libraries.
+- Do not add the key to **Cloud Agents → My Secrets** for this HTTP connection. Cursor proxies HTTP MCP requests through its backend and applies the saved headers there; the header is not read from the agent VM.
```
enesgules: Could not confirm this in Cursor's docs. Keep it only if you observed it directly; otherwise cut it.

docs/clients/cursor.mdx
```diff
@@ -19,6 +19,148 @@ For manual MCP installation or other configuration options, see [All MCP Clients
...
+
+<Frame>
+  <img src="/images/clients/cursor/cloud-agents/05-context7-json-config.jpg" alt="Context7 custom MCP JSON configuration in Cursor Cloud Agents" />
+</Frame>
+</Step>
+
+<Step title="Add and enable the server">
+Select **Add MCP**. Before starting the run, open **+** → **MCP Servers** again and confirm that `context7-api-key` is enabled.
+
+A green or ready status confirms that Cursor initialized the MCP server. It does not by itself prove that the request was authenticated.
```
enesgules: "A green or ready status" reads as two things. "A green status" is enough.

docs/clients/cursor.mdx
```diff
@@ -19,6 +19,148 @@ For manual MCP installation or other configuration options, see [All MCP Clients
...
+```text
+Use Context7 to query <YOUR_PRIVATE_LIBRARY_ID>. Summarize its installation docs.
+```
+
+If the private library can be read, the Cloud Agent is using the API key. If the agent reports **Monthly quota exceeded** even though the key has quota, the request is reaching Context7 anonymously; recheck the header name and start a new run after saving the MCP server.
+
+After the API-key header is applied, the same Cloud Agent should be able to resolve a library and return documentation without the anonymous quota error:
+
+<Frame>
+  <img src="/images/clients/cursor/cloud-agents/06-verify-context7.jpg" alt="Successful Context7 documentation query from a Cursor Cloud Agent" />
```
enesgules: This screenshot is about half empty canvas. Crop it.

docs/clients/cursor.mdx
```diff
@@ -19,6 +19,148 @@ For manual MCP installation or other configuration options, see [All MCP Clients
 
 ---
 
+## Cursor Cloud Agents
```
enesgules, line 85: Placement: a 140-line section now sits between the one-command install and "Setting Up Rules". Consider moving it below "Using Context7" so the common path stays short.

enesgules (commented):
Content is accurate and useful. Verified: screenshots contain no real key, both Cursor doc links and anchors resolve, live test against `mcp.context7.com/mcp` confirms the `Context7-API-Key` header is accepted. Two things to fix before approval (header name consistency, the Warning box), rest is minor.

Also: the PR summary says the guide is linked from the all-clients Cursor section, but the diff has no change to `docs/resources/all-clients.mdx`. Add the link or drop the claim.

docs/clients/cursor.mdx
fahreddinozcan: Addressed in 06a4256. Added the rationale: the distinct name avoids a collision with the installed OAuth plugin, while context7 is fine when the plugin is not installed.

docs/clients/cursor.mdx
fahreddinozcan: Addressed in 06a4256. Replaced the image with a tightly cropped HD PNG focused on the successful result.

docs/clients/cursor.mdx
fahreddinozcan: Addressed in 06a4256. Changed this to “A green status”.

docs/clients/cursor.mdx
fahreddinozcan: Addressed in 06a4256. Removed the unverified Cloud Agents → My Secrets claim.

docs/clients/cursor.mdx
fahreddinozcan: Addressed in 06a4256. The guide now states that both Authorization: Bearer and Context7-API-Key are accepted, and explains why the verified Cursor Cloud example uses the latter.

docs/clients/cursor.mdx
fahreddinozcan: Addressed in 06a4256. The warning now correctly says to use a hyphenated header because underscore-containing headers can be dropped by Cursor's proxy; it also clarifies that CONTEXT7_API_KEY remains valid for local stdio.

docs/clients/cursor.mdx
fahreddinozcan: Addressed in 06a4256. Moved the Cloud Agents guide below “Using Context7” so the common setup and usage path remains first.

enesgules (approved):

PR #3194, opened 2026-09-14
CTX7-2768: Disable unused MCP subscriptions

fahreddinozcan:
## Summary

- advertise static MCP tool, prompt, and resource capabilities so clients do not open unused notification subscriptions
- disable HTTP and stdio subscriptions by default from the shared subscription configuration helper
- treat explicit zero and invalid subscription limits as disabled while retaining positive values for deliberate opt-in deployments
- add a patch changeset for `@upstash/context7-mcp`

## Production context

The `MCPGatewayConnectionsHigh` alert showed about 63.5k active subscriptions globally and about 12.7k upstream connections on each of five Envoy gateway pods. Subscription p95 lifetime was exactly 300 seconds, matching the gateway stream-idle timeout.

Context7 does not publish tool, prompt, or resource change notifications, so these SSE streams consume gateway connections and memory without providing functionality.

Linear: https://linear.app/upstash/issue/CTX7-2768/disable-unused-mcp-notification-subscriptions

## Validation

- existing MCP test suite — 124 passed
- `pnpm --filter @upstash/context7-mcp lint:check`
- Prettier check
- Changeset status validation


ramazanoacar (approved):

ramazanoacar:
## Non-blocking suggestions

1. **Move the "unset → 0" decision into `getMaxSubscriptions`.** Right now the same ternary is duplicated at both call sites and the helper still returns 16000 for `undefined`, so the helper's tests (`defaults to 16000 subscriptions`) no longer describe production behaviour. Returning `0` from the helper when unset gives one source of truth and lets you cover the new default with a one-line test.
2. **`MCP_MAX_SUBSCRIPTIONS=0` (or any invalid value) currently *enables* subscriptions at 16000** via the fallback-with-warning path. That's the opposite of what an operator writing `=0` would expect now that "off" is the default. Treating `0`/invalid as disabled would be safer; at minimum the `.env.example` comment could say "leave unset to disable" so nobody reaches for `0`.
3. **Heads-up on log volume.** On the HTTP path a refused listen also calls `onerror`, which we route to `console.error`. Any non-SDK client that still sends `subscriptions/listen` unconditionally will now produce one error line per session. Probably fine, but worth watching after rollout, and easy to filter in `onerror` if it gets noisy.

fahreddinozcan:
Addressed the configuration suggestions in `00c2c21`:

- `getMaxSubscriptions` is now the single source of truth and defaults to `0`
- explicit `MCP_MAX_SUBSCRIPTIONS=0` is accepted as disabled
- invalid values warn and safely fall back to `0`
- duplicated call-site ternaries were removed

I left the logging behavior unchanged; clients honoring the advertised static capabilities will not send `subscriptions/listen`.

ramazanoacar (approved):

PR #3199, opened 2026-09-15
CTX7-2792: Document native AWS Bedrock setup

fahreddinozcan:
## Summary

- add a native AWS Bedrock setup guide for Context7 Enterprise
- document IAM permissions, workload roles, environment credentials, shared profiles, and Bedrock bearer keys
- document Claude and Titan model selection, cross-region inference profiles, and PrivateLink endpoints
- clarify credential precedence, offline-license networking, and data-processing behavior
- link the guide from Enterprise and Kubernetes documentation

## Validation

- docs navigation JSON parsed successfully
- MDX paths and internal links verified
- whitespace and patch checks passed

Linear: https://linear.app/upstash/issue/CTX7-2792/document-native-aws-bedrock-setup-for-context7-enterprise


ramazanoacar (commented):
Nice guide. The credential-precedence warning and the "offline license still needs AWS egress" note are exactly the things people trip on, glad they're in.

### Before merge

1. **IAM wording for inference profiles** (`aws-bedrock.mdx`, Prerequisites)
   "for those models **or** inference profiles" reads as either/or. For a `us.` / `eu.` / `apac.` profile, AWS requires `InvokeModel` on the profile ARN **and** on the foundation-model ARN in every destination region, otherwise you get `AccessDeniedException` even though the profile is allowed. Suggest: "for the models you use. For a cross-region inference profile, allow both the profile ARN and the underlying foundation-model ARN in each destination region." Linking the AWS geographic cross-region inference page would help.

2. **Changelog and version gating**
   There is no Bedrock entry in `changelog.mdx`, and the guide doesn't say which image version ships the provider. Since docs publish from master, could we add the changelog line and an "available from vX.Y" note, or hold the merge until the image is out?

### Optional

- **Field naming.** The guide calls it the "Bedrock endpoint" field, while `on-premise.mdx` maps it to "LLM Base URL". Worth aligning to whatever the UI shows.
- **Short-term Bedrock API keys expire after 12 hours.** Since the tab suggests pasting a key into the settings field, a one-line note that only long-term keys suit that path would save someone a confused morning.
- **Shared profile tab.** The mounted `~/.aws` files are usually 0600 on the host. If the container runs with `--user` or `runAsUser`, they need to be readable by that UID. One sentence would do.
- **Cross-region prefixes.** `global.` also exists now, so "such as `us.`, `eu.`, `apac.`" is safer than listing three.
- **Nav placement.** Integrations currently holds source connectors only. Bedrock at the top of that list reads a little odd. Either place it last or consider an "AI Providers" group later. Not worth blocking on.
- **On-premise intro.** Dropping "never leave your environment" is the honest call. The new description is a bit flat though. Something like "Your code, docs, and embeddings stay in your infrastructure; only the content needed for parsing and embedding goes to the AI provider you choose" keeps the point without over-promising.

fahreddinozcan:
Addressed the documentation review in 62b594b: clarified that cross-region profiles require permissions for the profile ARN plus source and destination foundation-model ARNs, linked the AWS geographic cross-region guide, added the v1.3.9 availability note and changelog entry, documented short-term API-key expiry and mounted-profile file permissions, included global profiles, aligned endpoint field names with the UI, improved the on-premise positioning, and moved Bedrock to the end of the Integrations list. Validation: docs.json parses, internal navigation remains valid, and all changed files pass Prettier. Ready for re-review; this docs PR should still merge after parser PR 746 and the v1.3.9 image release.

ramazanoacar (approved):
### Merge timing

You noted this should land after parser PR 746 and the v1.3.9 image release. Agreed. Two small things tied to that:

- The changelog entry is dated 2026-09-16. If the image ships on a different day, bump the label to match.
- Since nothing in the repo blocks an early merge, a "do not merge" label or draft status until v1.3.9 is out would keep someone from merging it by accident.

### Optional nit

- **Global profiles.** The IAM bullet covers geographic profiles, but a `global.` profile also needs the global foundation-model ARN (the one without region and account). Since the warning box now mentions `global.`, a short "global profiles have their own policy shape, see [link]" pointing at the AWS global cross-region page would close the loop.

PR #3200, opened 2026-09-15
CTX7-2681: Clarify CLI login recovery and share the API URL

enesgules:
When `ctx7 whoami` cannot check a saved session, it currently recommends logout and login for every failure. This can cause users to discard valid credentials during a server outage or network failure.

This change gives logout/login instructions only when the identity endpoint returns HTTP 401. Other HTTP errors, network failures, and invalid JSON report that the session could not be verified and ask the user to try again later. Login also uses the existing shared API URL. A patch changeset includes both changes.

Validation:

- All 372 CLI tests pass, including HTTP 401, HTTP 403, HTTP 500, rejected fetch, and invalid JSON cases.
- CLI lint, type checks, formatting, and build pass.
- The built CLI passes a local HTTP smoke test for those failures and a successful identity response. Saved access and refresh credentials remain unchanged in every case.


enesgules:
Additional verification: the built ctx7 binary passed a local HTTP smoke test for a rejected session, the logout-then-login hint, logout, device-code login, and whoami. The custom API URL handled every request. Filesystem permissions confined credential writes to the test directory.

ramazanoacar (approved):
One suggestion below, non-blocking.

The URL dedup is clean: the only reader of the old module-level URL was `performLogin`, whoami and the refresh path already went through `getBaseUrl()`, and the `preAction` hook still sets it before any command runs. Changeset and tests look right, CI is green.

### Suggestion

**Only give the logout-then-login hint on an auth failure** (`auth.ts`, `whoamiCommand` catch block)

`fetchWhoami` throws on any non-OK response and on network errors, and the catch block can't tell them apart. So a DNS failure, a corporate proxy, or a 5xx now tells the user to run `ctx7 logout`, which throws away a session that is probably still fine. For pre-0.5 OAuth users that also discards the refresh token.

Cheap fix: have `fetchWhoami` surface the status (or throw a small typed error), then:
- 401/403: "Session was rejected by the server. Run 'ctx7 logout', then 'ctx7 login'."
- anything else: "Couldn't reach Context7 to verify the session." (`describeConnectionError` in `utils/auth.ts` already produces nice text for the network case.)

Side benefit: the `status: 401` you added to the test would then actually drive a branch. Today the assertion passes for any failure, so the test name promises a little more than it checks.

Happy to approve as is and take this as a follow-up if you'd rather keep the PR small.

fahreddinozcan (changes requested):
The shared API URL cleanup is good, but the recovery path still conflates every `whoami` failure with a revoked session.

`fetchWhoami` throws the same error for every non-2xx response, and `whoamiCommand` also catches network failures and JSON parsing failures. As a result, an outage, timeout, malformed response, or 500 tells the user to run `ctx7 logout` and discard otherwise valid credentials. The new test supplies status 401, but the implementation never checks that status.

Please introduce a typed/status-aware unauthorized error and recommend logout/login only for 401 or the explicitly supported authorization status. Transient server and network failures should preserve credentials and report that identity could not be checked. Add coverage for at least a 500 response and a rejected fetch so the recovery message cannot regress back into this catch-all branch.

fahreddinozcan (approved):
Verified the updated implementation against the requested changes. The CLI now distinguishes an HTTP 401 from transient HTTP, network, and response-parsing failures; only the rejected session path recommends logout/login. The added 403, 500, rejected-fetch, and invalid-JSON tests close the catch-all regression. The shared base URL cleanup remains direct and removes duplicate auth state. Approved.

Issue #746, opened 2025-10-05
CodeGeex - connection closed issue


abdelhameedhamdy:
Hey there!,

When adding context7 MCP in codegeex mcp server:
```json
{
  "mcpServers": {
    "context7": {
      "command": "npx",
      "args": [
        "-y",
        "@upstash/context7-mcp",
        "--api-key",
        "xxx"
      ]
    }
  }
}
```
It fails to connect and gives `Connection closed`.

Any thoughts how to fix it!

Thanks

enesgules:
Does it support remote mcp servers? You could try connecting with http transport

abdelhameedhamdy:
> Does it support remote mcp servers? You could try connecting with http transport

Tried : 
```
{
  "mcpServers": {
    "context7": {
      "type": "http",
      "url": "https://mcp.context7.com/mcp",
      "headers": {
        "CONTEXT7_API_KEY": "xxx"
      }
    }
  }
}
```
got red error:
`Error Message
连接失败，'command'`

enesgules:
I know nothing about this MCP client sorry

abdelhameedhamdy:
update it works with the following config:
```json
{
  "mcpServers": {
    "context7": {
      "command": "/opt/homebrew/bin/node",
      "args": [
        "/opt/homebrew/lib/node_modules/@upstash/context7-mcp/dist/index.js"
      ],
      "env": {
        "CONTEXT7_API_KEY": "xxx"
      }
    }
  }
}
```

abdelhameedhamdy:
When trying to connect to it got the following error :
[错误] 程序运行报错，错误原因: MCP error -32602: Tool resolve_library_id not found

enesgules:
Hey @abdelhameedhamdy any updates on this? did you find a fix?

abdelhameedhamdy:
> Hey [@abdelhameedhamdy](https://github.com/abdelhameedhamdy) any updates on this? did you find a fix?

Same error, no fix till now.

enesgules:
Could you share the docs of this mcp client please, it seems to not be super popular

abdelhameedhamdy:
https://codegeex.cn/en-US
https://zhipu-ai.feishu.cn/wiki/CuvxwUDDqiErQUkFO2Tc4walnZY

abdelhameedhamdy:
> Could you share the docs of this mcp client please, it seems to not be super popular

[...]

PR #3206, opened 2026-09-16
CTX7-2807: Clarify Bedrock regional setup

fahreddinozcan:
## Summary

- clarify that native Bedrock requires the v1.3.9+ enterprise image
- document US, EU, and APAC inference-profile prefixes
- document Bedrock's supported environment, shared-profile, and workload-role credential chain

## Validation

- paired implementation and guide independently reviewed
- `docs.json` parses
- MDX formatting and internal links validated
- `git diff --check`

Merge after the CTX7-2804 parser PR.

Linear: CTX7-2807


ramazanoacar (approved):
Please merge after upstash/context7parser#750 as noted in the description.

Checked the claims against the parser branch:

- Version gating is consistent: the latest tag is v1.3.8, and both #746 (native Bedrock) and #750 (region-aware defaults) land in v1.3.9, so "v1.3.9 and later" is right for both statements.
- The inference-profile table matches the code: `us.`, `eu.`, `apac.` are exactly what `bedrockInferenceProfilePrefix` produces.
- The on-premise note on Bedrock authentication matches `BedrockProvider.ts`: `AWS_BEARER_TOKEN_BEDROCK`, environment credentials, shared profile, and workload role via the default credential chain.

One optional nit: "Updating these docs or configuration alone does not add Bedrock support to an earlier image" reads a bit oddly inside the docs themselves. The version requirement sentence on its own already makes the point. Fine to leave as is.

PR #3210, opened 2026-09-17
CTX7-2818: Document the Search API

fahreddinozcan:
## Summary

- document GET /api/v3/search in the public OpenAPI reference
- explain the one-request search flow, library hints, version, language, and response formats
- add the TypeScript SDK Search command page and navigation entry
- include examples for direct API and SDK usage
- preserve the existing two-step library search and context example
- document the SDK's complete JSON response, including optional rules

## Validation

- parsed docs/openapi.json as JSON
- ran Prettier on the changed documentation
- ran git diff --check
- verified that new documentation does not contain em dashes

## Dependency

This documents the endpoint introduced by https://github.com/upstash/context7app/pull/1163.


ramazanoacar (approved):
Good structure: the one-request flow up front, the two-step flow kept for when you need to pick the source, and the "add hints only when useful" guidance is the right framing. Mintlify deploy is green and the reference page is auto-generated from `openapi.json`, so no nav entry is needed.

### Before merge

1. **Error responses don't match the route** (`openapi.json`, `/v2/search`). The route returns 400, 401, 402, 404, 429 and 503 (it maps internal failures to 503, not 500), and the upstream spec in context7app#1163 lists exactly those. Suggest dropping 500 and 403 to match.

2. **Document the 404 semantics.** The generic `NotFoundError` examples talk about library IDs and tags, but this endpoint returns 404 `no_documentation_found` when nothing matched the question or hints. That's the case callers will hit most. A dedicated example on the 404 response, plus a sentence in the api-guide section ("a search with no matches returns 404, treat it as an empty result"), would cover it. Same for `search_failed` on 503.

### Optional

- **Version pattern is case-sensitive.** The route validates with the `i` flag, so `1.0.0-RC1` is accepted, but the documented pattern `[a-z0-9.-]` rejects it. `[a-zA-Z0-9.-]` matches upstream.
- **Default format.** The SDK page says text is the default. If #3211 switches to JSON, this page and the README example need the same change.
- **Merge order.** Docs publish on merge, so this should land after context7app#1163 is deployed.
- **Text example** (`"Library: /vercel/next.js\n\n### ..."`): I couldn't confirm the actual text layout from the upstream code, so worth a quick check against a real response before merge.

enesgules (commented):
## Review at `8d456ff`

The API reference link resolves: the summary "Search documentation" produces `/api-reference/search/search-documentation`. The "one request per call" statement matches the quota code. Four items:

### 1. The two OpenAPI copies now differ

On master, `/v2/context` and `/v2/libs/search` are identical in `docs/openapi.json` and the app's `public/openapi.json`. For `/v2/search` the summary, description, parameter layout (component refs here, inline there), response schema (plain object here, `allOf` with `ContextResponse` there), and response codes all differ. Pick one as the source and copy it to the other.

### 2. Version text is too soft

`docs/sdks/ts/commands/search.mdx` says "A version to prefer". With one library hint the server is strict and returns 404 when no matching tag exists. The page's own example uses one library with `15.4.0`. State the strict behavior, as the OpenAPI `VersionParam` description already partly does.

### 3. Error responses do not fit this endpoint

The 404 reuses the shared `NotFoundError` examples (`library_not_found`, `tag_not_found`, ...). Search returns `no_documentation_found`. The spec also lists 403 and 500, which the route does not produce; it returns 503 `search_failed` on failure.

### 4. SDK page has no error section

Add one line: a search with no match throws `Context7Error` with a 404 status.

### Dependency

This documents `language`, which is the open blocker on upstash/context7app#1163. Recheck if the fix there changes or removes the parameter. Merge only after that PR is deployed, because `/api/v2/search` currently serves the legacy library search.

fahreddinozcan:
Addressed the docs review in `d64badc`. `/v2/search` and `SearchContextResponse` now match the endpoint OpenAPI copy exactly, including the route's 400/401/402/404/429/503 responses and dedicated `no_documentation_found`/`search_failed` examples. The guide explains 404/503; the SDK page explains strict single-library versions, 404 `Context7Error`, and the SDK's JSON default versus the REST text default. Formatting and JSON validation pass. This PR should merge after the endpoint is deployed.

enesgules (approved):
Verified `d64badc`. All four items are resolved:

1. `/v2/search`, its parameters, and the three Search schemas are now identical to the app's `public/openapi.json`.
2. The SDK page states the strict single-library version rule.
3. 404 and 503 use the endpoint's own `no_documentation_found` and `search_failed` examples.
4. The SDK page documents the 404 `Context7Error`, and the guide explains 404 and 503.

Approving. Merge only after upstash/context7app#1163 is deployed, because `/api/v2/search` currently serves the legacy library search.

fahreddinozcan:
The endpoint moved to `GET /api/v3/search` because `/api/v2/search` still has callers of the library-search response (upstash/context7app#1163). The API guide examples and the OpenAPI path are updated; the path definition is identical to the app spec.

PR #3211, opened 2026-09-17
CTX7-2816: Add Search API to TypeScript SDK

fahreddinozcan:
## Summary

- add client.search(query, options) with text output by default and the complete typed JSON response
- support up to four library hints plus version and language preferences
- preserve source library IDs and optional global and library-specific rules in JSON output
- return the API response directly, without a lossy snippet conversion
- support repeated query parameters in the shared HTTP client
- add package documentation and a minor changeset

## Validation

- pnpm --filter @upstash/context7-sdk test
- pnpm --filter @upstash/context7-sdk typecheck
- pnpm --filter @upstash/context7-sdk lint:check
- pnpm --filter @upstash/context7-sdk build

All 47 SDK tests pass.

## Dependency

This client method calls the endpoint introduced by https://github.com/upstash/context7app/pull/1163.


ramazanoacar (approved):
One decision to settle before merge, the rest is optional.

Nice and tidy, this follows the existing command shape well, and returning the API response as-is instead of the lossy `Documentation[]` mapping is the right call for a multi-library result. The repeated-param support in `buildUrl` is minimal and covered by a test.

### Before merge

1. **Default format: `txt` here, `json` everywhere else.** `getContext` and `searchLibrary` both default to JSON with typed returns, so `client.search("q")` returning a string will surprise people who use the other two. The endpoint defaults to `txt`, but the SDK already diverges from the endpoint defaults for the other methods. I'd lean to `json` for consistency, but either is defensible. Worth deciding now since it's public API and a minor bump, and the docs PR (#3210) mirrors whatever we pick.

2. **Hold merge until the endpoint is live.** The changeset will publish on the next release, and the app PR (context7app#1163) is still open.

### Optional

- **No results throws.** The route returns 404 `no_documentation_found` when nothing matches, so a plain "nothing found" surfaces as a thrown `Context7Error` rather than empty arrays. That's fine, but a line in the `search` JSDoc would save users a try/catch surprise. Same for `search_failed` (503), which the default retry policy will retry.
- **`QueryParams` alias** in `commands/types.ts` is exported but unused anywhere. Widening it keeps it in sync with `Context7Request.query`, but deleting it or having `http/types.ts` reference it would remove the duplication.
- **Client-side guard for `version` without `libraries`.** `SearchLibraryCommand` already validates required args locally. A `Context7Error` before the request would match that pattern and give a clearer message than the 400. Not important.
- **Integration test.** Once the endpoint is deployed, a `search` case in `client.integration.test.ts` alongside the others would be nice.

enesgules (commented):
## Review at `4afe663`

Verified in a clean worktree: 47/47 tests pass, type-check and lint are clean. The design is minimal and correct: `client.search` returns the API response directly, the overloads type `txt` as `string` and `json` as `SearchResponse`, and the HTTP client repeats array query parameters.

### 1. Release order

Production `/api/v2/search` is still the legacy library search. It ignores `type` and returns `{ results, searchFilterApplied }`. If this SDK version ships before upstash/context7app#1163 is deployed, `client.search()` returns that object typed as `string`. Publish the changeset only after the app deploy.

### 2. "No result" throws

The endpoint returns 404 `no_documentation_found` when nothing matches, so `client.search()` throws `Context7Error`. That is acceptable, but it needs one line in the SDK docs page (noted on #3210).

### 3. Dependency on `language`

`language` is the open blocker on upstash/context7app#1163. The name and shape will probably stay. Recheck this PR if the fix there changes or removes the parameter.

No client-side validation for more than four libraries or a version without a library is needed. The server returns 400, which is the right place.

fahreddinozcan:
Addressed the default-format review in `a90273a`: `client.search()` now requests and returns structured JSON by default, consistent with `getContext` and `searchLibrary`; `{ type: "txt" }` returns text. The JSDoc notes 404 `Context7Error`. SDK tests (48/48), typecheck, and lint pass. This PR should publish only after the endpoint deploys.

enesgules (approved):
Verified `a90273a` in a clean worktree: 48/48 tests pass, type-check and lint are clean. The JSON default matches `getContext`, the overloads are correct for both formats, and the JSDoc states the 404 throw. The 503 retry statement in the docs is accurate: the HTTP client retries transient statuses on GET.

Approving. Publish the changeset only after upstash/context7app#1163 is deployed.

fahreddinozcan:
The endpoint moved to `GET /api/v3/search` because `/api/v2/search` still has callers of the library-search response (upstash/context7app#1163). `SearchCommand` and the endpoint list now use `v3/search`; the request-URL test is updated. Typecheck, lint, and 48/48 tests pass.

PR #3215, opened 2026-09-18
CTX7-2681: Document Docs7 login and deployment commands

enesgules:
## Summary

- Add one Docs7 CLI reference for login, logout, whoami, local preview, and deploy.
- Explain production by default, same-site updates, preview and slug options, upload exclusions, plan limits, and optional GitHub connection.
- Add CLI publishing to Quickstart and correct the GitHub-only statements in Builds and Intro.
- Keep this docs-only: five files, no application code, API schema, dependencies, or tests.

## Release dependency

Merge these docs after upstash/docs7#305 and upstash/context7app#1110 are released, including the npm package with the deploy command. Login is already merged. The new command examples describe the deploy PRs, not the current npm release.

## Validation

- Checked commands and options against CLI source and help at upstash/docs7@26113b3.
- All four changed MDX pages parse and have valid frontmatter.
- Navigation JSON, all 37 Docs7 navigation targets, and seven internal links with anchors pass.
- New CLI page passes Prettier. Existing docs formatting is preserved; docs are excluded from the repository formatter, and the forced docs.json formatting check also fails on master.
- git diff --check passes.
- Docs7 Preview build passed. All four changed pages return HTTP 200, and the CLI page contains the new command sections.

No user site was deployed and no account settings changed during validation.


ramazanoacar (approved):
Approve, with one fix before merge. Hold the merge until docs7#305 and context7app#1110 ship, as you noted.

Really solid page. I checked the commands, flags, defaults, credential paths, XDG rule, `CONTEXT7_API_KEY` precedence, name limit, `.docs7/project.json` behavior, `--slug` restriction, upload exclusions, and the Free/Pro rules against the CLI source and the app PR. All accurate. The GitHub-only corrections in Builds and Intro are the right call too.

### Before merge

**Downgrade recovery is missing a step** (`cli.mdx`, last Warning). "Deploy it again after the downgrade" won't work as written. Once the site is removed, `.docs7/project.json` still holds the old site ID, the CLI sends it, and the API answers 404 "The Docs7 site was not found." The app's own downgrade modal tells the user to delete `.docs7/project.json` first, then run `docs7 deploy`. The docs should say the same.

### Optional

- **All dot-prefixed paths are excluded from uploads**, not only "common private paths". The archiver skips every hidden file and directory except `.docs7ignore`, `.mintignore`, and `.gitignore`. Anyone with a dot-prefixed assets folder will wonder where it went, so one sentence under "Files in the upload" would help. The size limits (100,000 files, 1 GiB of source, 200 MiB compressed) could sit in the same paragraph.
- **"The CLI will no longer deploy to it"** after connecting GitHub: I couldn't find the message the CLI prints in that case, so worth confirming the wording against a real run.
- **Concurrent deploys.** The CLI takes a lock per directory and refuses a second `docs7 deploy` while one runs. Not essential, but CI users might hit it.

Issue #1216, opened 2025-12-12
Refresh request for /websites/elevenlabs_io


simon-vincent:
Please refresh the library /websites/elevenlabs_io as it is very large (1,560,578 tokens) and cannot be refreshed automatically.

Issue #355, opened 2025-06-30
Stuck in "The library is being processed" for over 14 hours when adding Datasaur documentation via website input


Yosua1011:
I am encountering an issue when attempting to add the Datasaur documentation via the **"Add from a Website"** functionality on the Context7 MCP Server.

- **Submitted path:** `https://docs.datasaur.ai/`  
- **Submission time:** Monday, June 30, 2025, at approximately 11:00 AM (GMT+7)  
- **Current status (as of July 1, 2025):** The interface continues to display the message:  
  `"The library is being processed"`

Based on the console output, it appears the process may be stalled. The last visible logs are as follows:

```
Chunking complete. Total chunks: 6.
Split into 4 potential sections.
Large but empty file: compatibility-and-updates/release-notes/version-6/6.36.md
Chunking complete. Total chunks: 6.
Split into 4 potential sections.
Starting chunking. Content length: 26457, Max chunk size: 5000
File compatibility-and-updates/release-notes/version-6/6.39.md is large (26457 chars), processing in chunks
Cache miss for: /context7/datasaur_ai###compatibility-and-updates/release-notes/version-6/6.39.md###95c8b52c2f47c4e246e8a753368d37d7
Split into 6 chunks
```

The status has not changed for over 14 hours, and no new logs have been generated since the above output. I have refreshed the page and rechecked multiple times, but the process appears to be stuck indefinitely.

I would appreciate it if you could investigate the issue and advise on how to proceed. If there is any additional information or context I can provide to assist in resolving this, please let me know.

Thank You.

enesakar:
refreshing

enesakar:
fixed: https://context7.com/context7/datasaur_ai 

I am closing this ticket. Please reopen it if you would like to follow up further.

ooopus:
> fixed: https://context7.com/context7/datasaur_ai
> 
> I am closing this ticket. Please reopen it if you would like to follow up further.

The Python 3.13 documentation

https://context7.com/add-library?tab=website&library=%2Fcontext7%2Fpython-3.13

is also encountering the same indexing issue. Could you take a look?

PR #3216, opened 2026-09-18
docs: clarify CLI site retention on Free

enesgules:
## Summary

- Explain that Free supports CLI deployment and that an owner can keep a CLI site when downgrading from Pro.
- Keep the site address, history, and saved CLI link. The current build stays online, and Free branding applies on the next deployment.
- Remove the old delete-and-deploy-again instructions.
- Remove numeric upload-size, file-count, and name-length limits from this guide. Server-side enforcement is unchanged.
- Keep the hidden-file exclusion guidance.

## Validation

MDX compilation, navigation, updated-copy checks, Prettier, and git diff --check passed.

Publish after https://github.com/upstash/docs7/pull/355 and https://github.com/upstash/context7app/pull/1216 are deployed.


fahreddinozcan (approved):
Follow-up review completed at c2d3d60. The documentation diff is clear and consistent with the intended behavior. Approved for content, with the PR stated release ordering: publish only after #355 and a corrected #1216 are deployed.

